PR #30, opened 2017-03-28
Add whitelist_python.py

### LINKED ISSUE #21: Whitelist defaultdict.default_factory
Originally reported by: **bitserver (Bitbucket: [bitserver](https://bitbucket.org/bitserver), GitHub: [bitserver](https://github.com/bitserver))**

----------------------------------------

I have a collections.defaultdict(). After I have finished populating it, I have to set its default_factory attribute to None. vulture then tells me that default_factory is an unused attribute, but I can't help it, as that's how defaultdict works.

----------------------------------------
- Bitbucket: https://bitbucket.org/jendrikseipp/vulture/issue/21

RJ722:
A standard whitelist file for Python/stdlib.

Closes #21


jendrikseipp:
The file looks good. Now we only need a test case under `tests/test_whitelist_python.py`. It should check that whitelist_python.py can be run by Python without errors and that code using the `default_factory` attribute is not marked as unused. You can look at `tests/test_script.py` to find out how to test Python files.

tests/test_whitelist_python.py
```diff
@@ -0,0 +1,28 @@
...
+    keywords for testing whitelist_python and check output.
+    """
+    example_file = 'tests/example.txt'
+    p = Popen(['vulture', example_file, 'whitelist_python.py'],
+            stdout=PIPE, stderr=PIPE)
+    output, err = p.communicate("")
+    print("Error: %s" % err)
+    print("Output: %s" % output)
+    if err.decode("utf-8").find("could not be found.") != -1:
+        raise FileNotFoundError("%s file wasn't found" % example_file)
```
RJ722: Use IOError because FileNotFoundError is not defined in python2.x

RJ722:
HI! @jendrikseipp 
Is the current approach in testing (adding an example file and running vulture on it) feasible?

jendrikseipp:
This is the only way to test whitelist files. However, normal code 
snippets should be tested in the same way as in `test_scavenging.py`. If 
you're adding multiple tests that belong together, you may consider 
adding a separate test file for that, like `test_format_strings.py`.

RJ722:
Thanks! :smile: 

Also, one thing I thought was necessary to bring to your attention was that I have named the example file with a `.txt` extension, because otherwise with the `.py`, it will cause the tests to fail as it would try to run vulture over example file without taking whitelist into consideration!

Also, do we need extra test_cases here?

whitelist_python.py
```diff
@@ -0,0 +1,3 @@
+import collections
+
+collections.defaultdict(list).default_factory
```
jendrikseipp: I propose to have both the example code and the whitelist code in one file, i.e., whitelist_python.py should be 

```
import collections

collections.defaultdict(list).default_factory = None
collections.defaultdict(list).default_factory
```

The .txt file can be removed. The test_whitelist_python_function() test should test that vulture finds no dead code in whitelist_python.py.

tests/test_whitelist_python.py
```diff
@@ -0,0 +1,20 @@
+import os.path
+import sys
+import subprocess
+
+from .test_script import call_vulture, REPO
+
+
+def test_whitelist_python():
```
jendrikseipp: Rename function to `test_whitelist_python_with_python`.

tests/test_whitelist_python.py
```diff
@@ -0,0 +1,20 @@
...
+    assert subprocess.call(
+        [sys.executable, 'whitelist_python.py'], cwd=REPO) == 0
+
+
+def test_whitelist_python_function():
```
jendrikseipp: Rename function to `test_whitelist_python_with_vulture`.

whitelist_python.py
RJ722: But, if we add `collections.defaultdict(list).defaultfactory`, then vulture won't show this as dead code even if we don't supply the whitelist file - as it would see that the user has not just left the deafult_factory set to None, but has used it later - contrary to what happened earlier, where it would report: ` whitelist_python.py:3: Unused attribute 'default_factory ` .

Am I wrong?

jendrikseipp: Consider the example code E = `defaultdict(list).default_factory = None`. What we want is a whitelist file W such that vulture makes the following report:

vulture(E + W) -> no dead code

It is ok to have the following (since it's the reason we need a whitelist file to begin with):

vulture(E) -> dead code

We always need to make sure that whitelist files don't contain unused code, but this is the case in my proposal. 

vulture(W) -> no dead code

We should add a third test for the last case.

RJ722: So, would we still need to separate E and W?

tests/example_whitelist_python.txt
```diff
@@ -1,5 +1,5 @@
-#  This file contains snippets of code which are currently in 
-#  whitelist_python.py which is used for testing. If changes are 
+#  This file contains snippets of code which are currently in
+#  whitelist_python.py which is used for testing. If changes are
```
jack17529: Please remove the spaces as they are unnecessary.

whitelist_python.py
jendrikseipp: We don't need to test this case.

jendrikseipp: OK, as said above, W should be

```
import collections

collections.defaultdict(list).default_factory = None
collections.defaultdict(list).default_factory
```

Then we need the following tests:

python(W) -> correct code
vulture(W) -> no dead code

RJ722:
Hi! @jendrikseipp 
Does this looks all right or do we need further enhancements?

jendrikseipp:
Looks good, thanks!

PR #34, opened 2017-03-30
Updated test_scavenging.py

### LINKED ISSUE #33: Mistake in calling function
When you assign a function to a variable you don't use the () but simply the name of the function.
In your case given def x(): ..., and variable silly_var you would do something like this:
silly_var = x
and then you can call the function either with
x()
or
silly_var()

In test_scavenging.py
func is written instead of func().

jack17529:
Mistake in calling function on my second read of code.

## Checklist:

Closes #33

### DIFF
--- a/tests/test_scavenging.py
+++ b/tests/test_scavenging.py
@@ -26,7 +26,7 @@ def test_function_object2():
 def func():
     pass
 
-func
+func()
 """)
     assert v.defined_funcs == ['func']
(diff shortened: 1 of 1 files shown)


jendrikseipp:
The code actually tested that a function is counted as "used" even if it is only referenced or assigned to another name wihtout being called. The code is therefore "correct" and doesn't need to be changed.

Issue #38, opened 2017-04-02
Use stdlib whitelist by default


RJ722:
Whitelists are a great way to tackle false positives, but we have to explicitly mention the whitelist fie every time we execute vulture, so **should vulture use whitelist by default?**

RJ722:
IMHO, we can let the user decide - we can have a *configuration file* - For now, it would contain only `whitelist: on/off`,  but it can be easily extended in future for further configuring vulture, if needed.

jendrikseipp:
I think the whitelists/stdlib.py file should be included by default. 
When we have other whitelist files, we can decide how to include them.

RJ722:
For extending `whitelists/stdlib.py` - can there be a definitive way of doing so?

What I had been thinking of was to **test it with major projects** (perhaps trending python projects on github) - this would serve us with two benefits:
* We can identify instances of what should ideally be in our whitelist file - as we may find any lesser known constructs (alike default_property itself).
* We would know of any bugs, if there, in vulture.

Is this approach feasible?

RJ722:
[coala/coala-bears/#1564](https://github.com/coala/coala-bears/issues/1564) will also get addressed through this issues.

jendrikseipp:
I think this is a great idea! Maybe we can even find projects that 
already use vulture or want to use vulture and can work on creating 
whitelists together.

RJ722:
Should we develop a script which downloads the trending project for the day, execute vulture on it and report the errors, if any!
This way we can ensure that we add something to the whitelist daily.
This would be fun and we can keep developing it - perhaps, may be extend it to e-mail the log to the project maintainer and he can get to us that way.

jendrikseipp:
I doubt that this can be automated. In any case, automation should not 
be the first step :-)

RJ722:
:+1: 

Already started with the [localstack](https://github.com/atlassian/localstack) repository - trending today! - [Here](https://pastebin.com/Gge1yp3q)'s the log file - I'll analyze the issues, convey them this notice and add anything suitable to whitelist tomorrow morning.

Don't you think that it is gonna be unmanageable if we continue this way, can you please suggest some better workflow through which we can track our perfomance!

jendrikseipp:
If you have SyntaxErrors like in the log, you're probably using the 
wrong Python version for calling vulture.

Something else to consider: If you're running vulture on a library, most 
functions will obviously be unused. In this case you should include the 
tests.

[...]

PR #50, opened 2017-06-13
Use whitelists/stdlib.py as default whitelist.

RJ722:
## Description

Append the location of `whitelists/stdlib.py` at the end of `args` passed, so that vulture consumes this whitelist in every run.
## Related Issue

Closes: https://github.com/jendrikseipp/vulture/issues/38
## Checklist:

### DIFF
--- a/tests/test_script.py
+++ b/tests/test_script.py
@@ -20,12 +20,16 @@ def get_output(args):
     return child.communicate()[0].decode("utf-8")
 
 
-def test_script_with_whitelist():
+def test_script_with_explicit_whitelist():
     assert call_vulture(['vulture/core.py', WHITELIST]) == 0
 
 
+def test_script_with_implicit_whitelist():
+    assert call_vulture(['vulture/core.py']) == 0
+
+
 def test_script_without_whitelist():
-    assert call_vulture(['vulture/core.py']) == 1
+    assert call_vulture(['vulture/core.py', '--exclude', 'whitelists']) == 1
 
 
 def test_exclude():
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -31,6 +31,7 @@ import codecs
 from fnmatch import fnmatchcase
 import optparse
 import os
+import pkgutil
 import re
 import sys
 import tokenize
@@ -166,15 +167,14 @@ class Vulture(ast.NodeVisitor):
         return modules
 
     def scavenge(self, paths):
-        modules = self._get_modules(paths)
-        included_modules = []
-        for module in modules:
-            if any(fnmatchcase(module, pattern) for pattern in self.exclude):
-                self.log('Excluded:', module)
-            else:
-                included_modules.append(module)
+        def exclude(name):
+            return any(fnmatchcase(name, pattern) for pattern in self.exclude)
 
-        for module in included_modules:
+        for module in self._get_modules(paths):
+            if exclude(module):
+                self.log('Excluded:', module)
+                continue
+
             self.log('Scanning:', module)
             try:
                 module_string = read_file(module)
@@ -183,6 +183,18 @@ class Vulture(ast.NodeVisitor):
                 print('You might want to change the encoding to UTF-8.')
             else:
                 self.scan(module_string, filename=module)
+
+            whitelist_names = ['stdlib.py']
+            for name in whitelist_names:
+                path = os.path.join('whitelists', name)
+                if exclude(path):
+                    self.log('Excluded Whitelist:', path)
+                else:
+                    module_data = pkgutil.get_data('vulture', path)
+                    if module_data is None:
+                        sys.exit('Error: Please use "python -m vulture.core".')
+                    module_string = module_data.decode("utf-8")
+                    self.scan(module_string, filename=path)
 
     def report(self):
         def file_lineno(item):
--- a/setup.py
+++ b/setup.py
@@ -51,4 +51,5 @@ setup(
     cmdclass={'test': PyTest},
     packages=['vulture'],
     package_dir={'vulture': 'vulture'},
+    package_data={'vulture': ['whitelists/*.py']},
 )


RJ722:
Having a default whitelist might have side-effects:
If we are whitelisting some dead code for a file, say `A`, and another file, say`B` contains the same dead code, we won't be able to report the unused code in `B`:

For example, in a file `dead.py`, I have:
```
import subprocess
import sys

def foo():
    pass

foo()
```

Only subprocess' import would be reported as unused because `stdlib` uses:
```
sys.stderr
sys.stdin
sys.stdout
```

This might be a bad user experience! :(

jendrikseipp:
Good catch. I changed the stdlib.py file to account for this and added a comment there.

vulture.py
```diff
@@ -394,6 +395,10 @@ def csv(option, _, value, parser):
 
 def main():
     options, args = parse_args()
+    whitelist_path = os.path.abspath(stdlib.__file__)
+    if whitelist_path.endswith('.pyc'):
+        whitelist_path = whitelist_path[:-1]
+    args.append(whitelist_path)
```
jendrikseipp: This solution only works when calling vulture from the command line, not as a library. You can add the following code into scavenge  

```
modules = self._get_modules(paths)
modules.append(_get_stdlib_whitelist_file())
```
and somewhere at the top of the script

```
def _get_stdlib_whitelist_file():
    script = os.path.abspath(__file__)
    whitelist_dir = os.path.dirname(script)
    return os.path.join(whitelist_dir, 'stdlib.py')
```

As a side effect, users can now disable default whitelist files by using `--exclude whitelists` since "whitelists" is part of the path name.

RJ722: > As a side effect, users can now disable default whitelist files by using --exclude whitelists since "whitelists" is part of the path name.

That's a win-win 🎉

RJ722:
Done making changes!

vulture.py
```diff
@@ -35,6 +35,8 @@
 import sys
 import tokenize
 
+from whitelists import stdlib
+
```
jendrikseipp: This should be obsolete.

vulture.py
```diff
@@ -165,8 +167,14 @@ def _get_modules(self, paths, toplevel=True):
                 sys.exit('Error: %s could not be found.' % path)
         return modules
 
+    def _get_stdlib_whitelist_file(self):
```
jendrikseipp: This should be a function, not a method, since it doesn't need access to the Vulture object.

vulture.py
```diff
@@ -165,8 +167,14 @@ def _get_modules(self, paths, toplevel=True):
...
+        script = os.path.abspath(stdlib.__file__)
```
jendrikseipp: use __file__ instead of stdlib.__file__

RJ722: Even if we get the location of the `vulture.py`, can we have a relative path to `stdlib`?

jendrikseipp: I think we need to change the way vulture is installed. Currently, only the main script is installed. Since no directory is created for vulture, we can't add the stdlib.py file anywhere. I'm not sure what the best solution is to this problem. I'll do some reading, maybe you have an idea?

https://docs.python.org/3/distutils/setupscript.html#installing-package-data
https://docs.python.org/3/library/pkgutil.html#pkgutil.get_data
https://stackoverflow.com/a/5899643

jendrikseipp: I think it's best to turn vulture into a package first, before we include the default whitelist. Can you open an issue for this? If you want to start a pull request, we need the following layout:

```
vulture/
    __init__.py (contains __version__)
    core.py (with the contents from old vulture.py)
```

vulture/core.py
```diff
@@ -176,8 +181,9 @@ def scavenge(self, paths):
 
         for module in included_modules:
             self.log('Scanning:', module)
+            module_string = _get_stdlib_whitelist()
             try:
-                module_string = read_file(module)
+                module_string += read_file(module)
```
RJ722: This seems to be highly inefficient!
Can we use `pkg_resources`

We can then change `_get_stdlib_whitelist` to:
```python
def _get_stdlib_whitelist():
    """
    Returns absolute path of the `stdlib` whitelist.
    """
    return pkg_resources.resource_filename('vulture', 'whitelists/stdlib.py')
```

jendrikseipp: I think it's better to use pkgutil.get_data. The change is more complicated than anticipated. Let me propose the following solution:

```
    def scavenge(self, paths):
        def exclude(name):
            return any(fnmatchcase(name, pattern) for pattern in self.exclude)

        for module in self._get_modules(paths):
            if exclude(module):
                self.log('Excluded:', module)
                continue

            self.log('Scanning:', module)
            try:
                module_string = read_file(module)
            except VultureInputException as err:
                print('Error: Could not read file %s - %s' % (module, err))
                print('You might want to change the encoding to UTF-8.')
            else:

        whitelist_names = ['stdlib.py']
        for name in whitelist_names:
            path = os.path.join('whitelists', name)
            if exclude(path):
                self.log('Excluded whitelist:', path)
            else:
                module_data = pkgutil.get_data('vulture', path)
                if module_data is None:
                    sys.exit('Error: Please use "python -m vulture.core".')
                module_string = module_data.decode("utf-8")
                self.scan(module_string, filename=path)
```

There are more changes needed afterwards: the tests have to use `python -m vulture.core` instead of `python vulture.py`.

vulture/core.py
```diff
@@ -87,6 +88,10 @@ def read_file(filename):
         raise VultureInputException(err)
 
 
+def _get_stdlib_whitelist():
+    return pkgutil.get_data('vulture', 'whitelists/stdlib.py').decode("UTF-8")
```
jendrikseipp: After you apply the change to `scavenge()` we don't need this function anymore.

tests/test_script.py
```diff
@@ -20,12 +20,12 @@ def get_output(args):
...
 
-def test_script_with_whitelist():
+def test_script_with_explicit_whitelist():
     assert call_vulture(['vulture/core.py', WHITELIST]) == 0
 
 
-def test_script_without_whitelist():
-    assert call_vulture(['vulture/core.py']) == 1
+def test_script_with_implicit_whitelist():
+    assert call_vulture(['vulture/core.py']) == 0
```
jendrikseipp: There should also be `test_script_without_whitelist()` which uses `--exclude whitelists/stdlib.py`.

vulture/core.py
```diff
@@ -184,6 +184,18 @@ def scavenge(self, paths):
...
 
+            whitelist_names = ['stdlib.py']
+            for name in whitelist_names:
+                path = os.path.join('whitelists', name)
+                if exclude(path):
+                    self.log('Excluded Whitelist:', path)
+                else:
+                    module_data = pkgutil.get_data('vulture', path)
+                    if module_data is None:
+                        sys.exit('Error: Please use "python -m vulture.core".')
```
RJ722: Thank You @jendrikseipp 😄 

Just one thing which is a little blurry at the moment: Why do we need this check and also wouldn't running `python -m vulture.core` just change the entrypoint?

jendrikseipp: Previously it was possible to run "python core.py". Now this is not possible anymore.

RJ722: How will it impact `pkgutil.get_data`?

Also, if I run `python vulture/core.py vulture/core.py --exclude whitelists/stdlib.py`, this seems to be working great.

RJ722:
Do we need to add any more tests?

vulture/core.py
jendrikseipp: I don't understand what you mean, but now everything should work fine :-)

Issue #48, opened 2017-06-07
Implement confidence values


RJ722:
It would be nice if vulture reports a confidence value associated with every result.

Strategy for calculating confidence values:
* `imports` are reported with 100% confidence
* Anything we are not sure about should be reported with 50% confidence
* `*` imports should with 0 % confidence

| Construct | Confidence |
|     :---:      |          :---:    |
|import	| 100%|
|from foo import *	|0%|
|variable	|50%|
|function	|50%|
|class	|50%|

jendrikseipp:
What's the status on this?

RJ722:
I am sorry for the late reply, I don't know how come this didn't show up in my notifications.

Well, I am done with the basic functionality, it's just I am trying to differentiate between `*` imports and normal imports.
I should raise a PR anyway 😛 .

RJ722:
Now, when I look into it, `vulture` doesn't reports `*` imports, so we don't need to do anything. Simply associate 100% confidence with `imports`. 😄

jendrikseipp:
I have thought about this some more and came to the conclusion that, at least for now, we should add the confidence values where they're actually needed, i.e. in the VultureBear.

RJ722:
Can we reopen this @jendrikseipp ?

Confidence values:
* Unreachable code - 1.0
* Unused import - 0.9
* other unused items - 0.6

jendrikseipp:
Sure, good idea. Now that we can detect unreachable code with 100% confidence, it makes sense to show these values in the output.

PR #57, opened 2017-06-23
Add confidence values

RJ722:
Report confidence values alongwith results

## Related Issue

Closes: #48 

## Checklist:

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -40,6 +40,11 @@ from vulture import utils
 
 __version__ = '0.22'
 
+CONFIDENCE_MAP = {
+    'unreachable_code': 1,
+    'import': 0.9
+}
+
 # The ast module in Python 2 trips over "coding" cookies, so strip them.
 ENCODING_REGEX = re.compile(
     r"^[ \t\v]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*?$", flags=re.M)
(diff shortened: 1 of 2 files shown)


vulture/core.py
```diff
@@ -191,8 +205,9 @@ def file_lineno(item):
         for item in sorted(
                 self.unused_funcs + self.unused_imports + self.unused_props +
                 self.unused_vars + self.unused_attrs, key=file_lineno):
-            print("%s:%d: Unused %s '%s'" % (
-                format_path(item.filename), item.lineno, item.typ, item))
+            print("%s:%d: Unused %s '%s' (%s%% sure)" % (
+                format_path(item.filename), item.lineno, item.typ, item,
+                item.confidence))
```
RJ722: I am not sure if we would want to format the output this way, does this looks good?

jendrikseipp:
I have thought about this some more and came to the conclusion that we should add the confidence values where they're actually needed, i.e. in the VultureBear.

Regarding the implementation, I think it should be an error to ask CONFIDENCE_MAP for a type it doesn't know about, i.e., use CONFIDENCE_MAP[typ] instead of CONFIDENCE_MAP.get(typ, 0).

RJ722:
commits from this PR can be merged through #86. Therefore, closing this.

Issue #59, opened 2017-06-25
Choose whitelists depending on used imports


RJ722:
Turn stdlib.py into "`stdlib`" dir with files "`ast.py`", "`sys.py`", etc. Then if "`ast`" is imported, use "`ast`" whitelist.

PR #63, opened 2017-07-02
Choose whitelist file based on imports

RJ722:
Related Issue

Closes: #59 

## Checklist:

### DIFF
--- a/tests/test_script.py
+++ b/tests/test_script.py
@@ -1,3 +1,4 @@
+import glob
 import os.path
 import subprocess
 import sys
@@ -6,7 +7,7 @@ from vulture import __version__
 
 DIR = os.path.dirname(os.path.abspath(__file__))
 REPO = os.path.dirname(DIR)
-WHITELIST = os.path.join(REPO, 'vulture', 'whitelists', 'stdlib.py')
+WHITELISTS = glob.glob(os.path.join(REPO, 'vulture', 'whitelists', '*.py'))
 
 
 def call_vulture(args, **kwargs):
@@ -24,7 +25,8 @@ def get_output(args):
 
 
 def test_script_with_explicit_whitelist():
-    assert call_vulture(['vulture/core.py', WHITELIST]) == 0
+    for whitelist in WHITELISTS:
+        assert call_vulture(['vulture/core.py', whitelist]) == 0
 
 
 def test_script_with_implicit_whitelist():
@@ -48,11 +50,13 @@ def test_dir():
 
 
 def test_whitelist_with_python():
-    assert subprocess.call([sys.executable, WHITELIST], cwd=REPO) == 0
+    for whitelist in WHITELISTS:
+        assert subprocess.call([sys.executable, whitelist], cwd=REPO) == 0
 
 
 def test_whitelist_with_vulture():
-    assert call_vulture([WHITELIST]) == 0
+    for whitelist in WHITELISTS:
+        assert call_vulture([whitelist]) == 0
 
 
 def test_pyc():
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -184,13 +184,16 @@ class Vulture(ast.NodeVisitor):
             else:
                 self.scan(module_string, filename=module)
 
-            whitelist_names = ['stdlib.py']
-            for name in whitelist_names:
-                path = os.path.join('whitelists', name)
+            for name in self.defined_imports:
+                path = os.path.join('whitelists', name) + '.py'
                 if exclude(path):
                     self.log('Excluded Whitelist:', path)
                 else:
-                    module_data = pkgutil.get_data('vulture', path)
+                    try:
+                        module_data = pkgutil.get_data('vulture', path)
+                    except IOError:
+                        # Most imports don't have a whitelist.
+                        continue
                     if module_data is None:
                         sys.exit('Error: Please use "python -m vulture.core".')
                     module_string = module_data.decode("utf-8")
--- a/vulture/whitelists/ast.py
+++ b/vulture/whitelists/ast.py
@@ -1,0 +1,17 @@
+from whitelist_utils import Whitelist
+
+# NodeVisitor methods are called implicitly.
+# TODO: Add missing methods.
+whitelist_node_visitor = Whitelist()
+whitelist_node_visitor.visit_arg
+whitelist_node_visitor.visit_alias
+whitelist_node_visitor.visit_Assign
+whitelist_node_visitor.visit_Attribute
+whitelist_node_visitor.visit_ClassDef
+whitelist_node_visitor.visit_comprehension
+whitelist_node_visitor.visit_For
+whitelist_node_visitor.visit_FunctionDef
+whitelist_node_visitor.visit_Import
+whitelist_node_visitor.visit_ImportFrom
+whitelist_node_visitor.visit_Name
+whitelist_node_visitor.visit_Str
--- a/vulture/whitelists/collections.py
+++ b/vulture/whitelists/collections.py
@@ -1,0 +1,5 @@
+from whitelist_utils import Whitelist
+
+# To free memory, the "default_factory" attribute can be set to None.
+whitelist_defaultdict = Whitelist()
+whitelist_defaultdict.default_factory
--- a/vulture/whitelists/sys.py
+++ b/vulture/whitelists/sys.py
@@ -1,0 +1,7 @@
+from whitelist_utils import Whitelist
+
+# Never report redirected streams as unused.
+whitelist_sys = Whitelist()
+whitelist_sys.stderr
+whitelist_sys.stdin
+whitelist_sys.stdout
--- a/vulture/whitelists/whitelist_utils.py
+++ b/vulture/whitelists/whitelist_utils.py
@@ -29,28 +29,4 @@ class Whitelist:
         pass
 
 
-# NodeVisitor methods are called implicitly.
-# TODO: Add missing methods.
-whitelist_node_visitor = Whitelist()
-whitelist_node_visitor.visit_arg
-whitelist_node_visitor.visit_alias
-whitelist_node_visitor.visit_Assign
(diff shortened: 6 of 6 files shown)


vulture/core.py
```diff
@@ -88,6 +88,15 @@ def read_file(filename):
         raise VultureInputException(err)
 
 
+def _get_suitable_whitelists(defined_imports):
+    path = os.path.dirname(os.path.abspath(__file__))
```
jendrikseipp: path -> vulture_path

vulture/core.py
```diff
@@ -88,6 +88,15 @@ def read_file(filename):
...
+    whitelist_dir = os.path.join(path, 'whitelists')
+    all_whitelists = os.listdir(whitelist_dir)
```
RJ722: Is there any other way through which we can enlist the `package_data`? 
(There is nothing as such in the documentation) :(

vulture/core.py
```diff
@@ -88,6 +88,15 @@ def read_file(filename):
...
+    defined_imports = [imp + '.py' for imp in defined_imports]
+    whitelist_names = [imp for imp in defined_imports if imp in all_whitelists]
```
jendrikseipp: Turn defined_imports and whitelist_names into sets, compute their intersection, then return the sorted intersection, but with "whitelists/" prepended. This has the advantage that we don't need to know the dir name "whitelists" in two parts of the code.

vulture/whitelists/stdlib.py
```diff
@@ -29,28 +29,4 @@ def __getattr__(self, _):
...
-# To free memory, the "default_factory" attribute can be set to None.
-whitelist_defaultdict = Whitelist()
-whitelist_defaultdict.default_factory
-
-# Never report redirected streams as unused.
-whitelist_sys = Whitelist()
-whitelist_sys.stderr
-whitelist_sys.stdin
-whitelist_sys.stdout
+Whitelist
```
jendrikseipp: I think this file should be renamed to `whitelist_utils.py`

vulture/core.py
```diff
@@ -184,7 +193,7 @@ def exclude(name):
             else:
                 self.scan(module_string, filename=module)
 
-            whitelist_names = ['stdlib.py']
+            whitelist_names = _get_suitable_whitelists(self.defined_imports)
```
jendrikseipp: whitelist_names are now whitelist_paths

RJ722:
Should we use `pkg_resources.resource_listdir('vulture', 'whitelists')` instead? 
> resource_listdir(package_or_requirement, resource_name)
List the contents of the named resource directory, just like os.listdir except that it works even if the resource is in a zipfile.

Reference: http://setuptools.readthedocs.io/en/latest/pkg_resources.html#basic-resource-access

tests/test_script.py
```diff
@@ -6,7 +6,11 @@
 
 DIR = os.path.dirname(os.path.abspath(__file__))
 REPO = os.path.dirname(DIR)
-WHITELIST = os.path.join(REPO, 'vulture', 'whitelists', 'stdlib.py')
+WHITELIST_DIR = os.path.join(REPO, 'vulture', 'whitelists')
+WHITELISTS = []
+for name in os.listdir(WHITELIST_DIR):
+    if name.endswith('.py'):
+        WHITELISTS.append(os.path.join(WHITELIST_DIR, name))
```
RJ722: That looks kind of dirty.

Can we do this instead:
```python
WHITELIST_PATH = os.path.join(REPO, 'vulture', 'whitelists', '*.py')
WHITELISTS = glob.glob(WHITELIST_PATH)
```

jendrikseipp:
Globbing and os.listdir will not work when vulture is distributed in a zip file (e.g., as an egg or wheel). Using `pkg_resources.resource_listdir` is an option, but I'd rather not let the source code depend on setuptools. We could use pkgutil.iter_modules or pkgutil.walk_packages to list all modules, but the behavior differs across Python versions. I propose to use the easy solution and just try to import the modules that are actually used in the tested code and skip a module if it can't be imported (since there is no whitelist for that module).

jendrikseipp:
Then we can use `https://docs.python.org/2/library/inspect.html#inspect.getsource` for the found modules to obtain the source code.

vulture/core.py
```diff
@@ -90,12 +89,16 @@ def read_file(filename):
 
 
 def _get_suitable_whitelists(defined_imports):
-    all_whitelists = set(pkg_resources.resource_listdir('vulture',
-                                                        'whitelists'))
-    defined_imports = set([imp + '.py' for imp in defined_imports])
-    common = sorted(all_whitelists.intersection(defined_imports))
-    whitelist_paths = [os.path.join('whitelists', name) for name in common]
-    return whitelist_paths
+    wh_names = []
```
jendrikseipp: explicit is better than implicit: wh_names -> whitelist_names

vulture/core.py
```diff
@@ -90,12 +89,16 @@ def read_file(filename):
...
+    defined_imports = ['wh_' + imp for imp in defined_imports]
+    wh_mods = __import__('whitelists', globals(), locals(), defined_imports, 1)
```
jendrikseipp: It would be good if we could avoid prepending all whitelists with a prefix like "wh_". Can we import the modules one by one using something like `__import__('vulture.whitelists.whitelist_name', ...)`? If we prepend "vulture.whitelists", maybe we don't have to rename the files.

RJ722: > It would be good if we could avoid prepending all whitelists with a prefix like "wh_".

Okay. I did this to prevent confusions while debugging.
> Can we import the modules one by one using something like __import__('vulture.whitelists.whitelist_name', ...)?  If we prepend "vulture.whitelists", maybe we don't have to rename the files.

Is this required only so that we don't have to rename the files?
If yes, then we can just remove the line:
```python
defined_imports = ['wh_' + imp for imp in defined_imports]
```
and then we can rename the files back to 'sys.py', etc.

Let me make changes in the PR itself, so we can have better insights.

jendrikseipp: Thinking about this some more it seems a little strange to import a module only to convert it back to text. Maybe it's better to just use pkgutil.get_data(). If a defined import can't be loaded by pkgutil.get_data() the function returns None, and we try the next defined import.

RJ722: Indeed!

tests/test_script.py
```diff
@@ -6,7 +7,8 @@
 
 DIR = os.path.dirname(os.path.abspath(__file__))
 REPO = os.path.dirname(DIR)
-WHITELIST = os.path.join(REPO, 'vulture', 'whitelists', 'stdlib.py')
+WHITELIST_GLOB = os.path.join(REPO, 'vulture', 'whitelists', '*.py')
+WHITELISTS = glob.glob(WHITELIST_GLOB)
```
RJ722: We have the following options, for just the test part:

* os.listdir
* glob.glob
* pkg_resources.resource_listdir

Do we have other alternatives? Which one's the best in your opinion?
Thanks!

jendrikseipp: I'd prefer globbing. You can combine the two lines above. No need to keep a reference to the glob expression around.

vulture/core.py
```diff
@@ -184,13 +184,16 @@ def exclude(name):
             else:
                 self.scan(module_string, filename=module)
 
-            whitelist_names = ['stdlib.py']
-            for name in whitelist_names:
+            for name in self.defined_imports:
                 path = os.path.join('whitelists', name)
```
jendrikseipp: append ".py" already here.

vulture/core.py
```diff
@@ -184,13 +184,16 @@ def exclude(name):
...
                 if exclude(path):
                     self.log('Excluded Whitelist:', path)
                 else:
-                    module_data = pkgutil.get_data('vulture', path)
+                    try:
+                        path += '.py'
+                        module_data = pkgutil.get_data('vulture', path)
+                    except IOError:
```
jendrikseipp: why do we ignore IOErrors? Or is this raised when a file can't be found?

vulture/core.py
RJ722: Will the user prefer to give `.py` extension in the `--exclude` flag?

vulture/core.py
RJ722: It's the case where we don't find a whitelist.

vulture/core.py
jendrikseipp: She might, therefore we should add it. `--exclude sys` will ignore `sys.py` and `system.py`.

vulture/core.py
jendrikseipp: OK. Can you add the comment `// We asked for a non-existent file. The error is expected.` above `pass`, please?

RJ722: Done!

vulture/core.py
```diff
@@ -184,13 +184,18 @@ def exclude(name):
             else:
                 self.scan(module_string, filename=module)
 
-            whitelist_names = ['stdlib.py']
-            for name in whitelist_names:
+            for name in self.defined_imports:
                 path = os.path.join('whitelists', name)
+                path += '.py'
```
jendrikseipp: please put this on the previous line: `path = os.path.join('whitelists', name) + '.py'`

vulture/core.py
```diff
@@ -184,13 +184,18 @@ def exclude(name):
...
                 if exclude(path):
                     self.log('Excluded Whitelist:', path)
                 else:
-                    module_data = pkgutil.get_data('vulture', path)
+                    try:
+                        module_data = pkgutil.get_data('vulture', path)
+                    except IOError:
+                        # We asked for a non-existent file.
+                        # The error is expected.
```
jendrikseipp: to make this fit one line:

    # Most imports don't have a whitelist.

RJ722: Done! 😄

jendrikseipp:
Nice work!

RJ722:
Thank You ! 😄

Issue #61, opened 2017-06-29
Differentiate between functions and classes


RJ722:
Add a “defined_classes” list and store the defined classes there.

PR #64, opened 2017-07-05
Differentiate between functions and classes

RJ722:
Closes: #61 

## Checklist:

### DIFF
--- a/tests/test_scavenging.py
+++ b/tests/test_scavenging.py
@@ -75,8 +75,10 @@ def func():
 
 func()
 """)
-    assert v.unused_funcs == ['Bar']
-    assert v.defined_funcs == ['Bar', 'func', 'func']
+    assert v.defined_classes == ['Bar']
+    assert v.defined_funcs == ['func', 'func']
+    assert v.unused_classes == ['Bar']
+    assert v.unused_funcs == []
 
 
 def test_attribute1(v):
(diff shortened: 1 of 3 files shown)


vulture/core.py
```diff
@@ -206,7 +207,8 @@ def file_lineno(item):
         unused_item_found = False
         for item in sorted(
                 self.unused_funcs + self.unused_imports + self.unused_props +
-                self.unused_vars + self.unused_attrs, key=file_lineno):
+                self.unused_classes + self.unused_vars + self.unused_attrs,
+                key=file_lineno):
```
RJ722: Should we introduce a new variable `all = self.unused_funcs + self.unused_imports + self.unused_props + self.unused_classes + self.unused_vars + self.unused_attrs` ?

tests/test_scavenging.py
```diff
@@ -159,8 +163,9 @@ class Bar():
 b = Bar()
 """)
     assert v.used_attrs == []
-    assert v.defined_funcs == ['Bar']
-    assert v.unused_funcs == []
+    assert v.defined_classes == ['Bar']
+    assert v.unused_classes == []
+    assert v.unused_vars == ['b']
```
jendrikseipp: Nice catch!

vulture/core.py
jendrikseipp: But maybe this is useful for external tools like VultureBear. I'll add a method.

jendrikseipp:
Nice work!

RJ722:
Thank You! 😄

Issue #68, opened 2017-07-12
Use utils.py


RJ722:
IMHO, many things from `core` can be moved to `utils`, otherwise we may face circular imports.

Our utils may contain:
```py
ENCODING = 'utf-8'
ENCODING_REGEX = re.compile(
    r"^[ \t\v]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*?$", flags=re.M)

class VultureInputException(Exception):
    pass
```
We may also export ` _format_path` method as `format_path`:
```py
def format_path(path):
    if not path:
        return path
    relpath = os.path.relpath(path)
    return relpath if not relpath.startswith('..') else path
```
Also,
```py
def read_file(filename):
    # Python >= 3.2
    try:
        # Use encoding detected by tokenize.detect_encoding().
        with tokenize.open(filename) as f:
            return f.read()
    except (SyntaxError, UnicodeDecodeError) as err:
        raise VultureInputException(err)
    except AttributeError:
        # tokenize.open was added in Python 3.2.
        pass

    # Python < 3.2
    try:
        with codecs.open(filename, encoding=ENCODING) as f:
            return f.read()
    except UnicodeDecodeError as err:
        raise VultureInputException(err)


class LoggingList(list):
    def __init__(self, name, verbose):
        self._name = name
        self._verbose = verbose
        return list.__init__(self)

    def append(self, item):
        if self._verbose:
            print('{0} <- {1}'.format(self._name, item))
        list.append(self, item)

```

PR #69, opened 2017-07-12
Add utils.py

RJ722:
Closes: #68

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -27,25 +27,18 @@
 from __future__ import print_function
 
 import ast
-import codecs
 from fnmatch import fnmatchcase
 import optparse
 import os
 import pkgutil
 import re
 import sys
-import tokenize
 
+from vulture.utils import (ENCODING_REGEX, format_path, LoggingList, read_file,
+                           VultureInputException)
 from vulture.lines import estimate_lines
 
 __version__ = '0.15'
-
-# The ast module in Python 2 trips over "coding" cookies, so strip them.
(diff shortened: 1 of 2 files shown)


vulture/core.py
```diff
@@ -27,26 +27,19 @@
...
 import optparse
 import os
 import pkgutil
 import re
 import sys
-import tokenize
 
+from vulture.utils import (ENCODING_REGEX, format_path, LoggingList, read_file,
+                           VultureInputException)
 from vulture.lines import estimate_lines
```
jendrikseipp: lines before utils

vulture/utils.py
```diff
@@ -0,0 +1,54 @@
...
+import os
+import re
+import tokenize
+
+# Encoding to use when converting input files to unicode.
+ENCODING = 'utf-8'
+
+# The ast module in Python 2 trips over "coding" cookies, so strip them.
+ENCODING_REGEX = re.compile(
+    r"^[ \t\v]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+).*?$", flags=re.M)
```
jendrikseipp: I'd rather keep ENCODING_REGEX in core.py since it'll only be used in the scan() function.

vulture/core.py
jendrikseipp: I'd prefer to just `from vulture import lines` and on a separate line `from vulture import utils` and use e.g. `utils.format_path` in the code. This makes it clear where the function comes from on first sight.

PR #70, opened 2017-07-13
Traverse children nodes for complete source

RJ722:
Traverse children nodes in a sequential order to determine the source of the given node.

## Checklist:

### DIFF
--- a/tests/test_size.py
+++ b/tests/test_size.py
@@ -31,16 +31,6 @@ class Foo:
     check_size(example, 3)
 
 
-def test_size_return():
-    example = """
-class Foo:
-    a = 1
-    b = 2
-    return 3
-"""
-    check_size(example, 4)
-
-
 def test_size_class():
     example = """
 class Foo(object):
--- a/tests/test_size.py
+++ b/tests/test_size.py
@@ -1,15 +1,16 @@
 import ast
 
-from vulture.lines import count_lines
-
 import pytest
 
-HAS_ASYNC = getattr(ast, 'AsyncFunctionDef', None)
+from vulture.lines import count_lines
+
+HAS_ASYNC = hasattr(ast, 'AsyncFunctionDef')
 
 
-def skip_if_not_has_async(item):
-    if callable(item) and not HAS_ASYNC:
-            pytest.mark.skip(item, "needs async definitions")
+def skip_if_not_has_async(function):
+    if not HAS_ASYNC:
+        pytest.mark.skip(
+            function, reason="needs async support (added in Python 3.5)")
 
 
 def check_size(example, size):
@@ -172,19 +173,20 @@ class Foo:
     check_size(example, 3)
 
 
+# We currently cannot handle code ending with multiline statements.
 def test_size_multi_line_return():
     example = """
 class Foo:
-    def long_string_return(o):
+    def foo():
         return (
             'very'
             'long'
             'string')
 """
-    # We currently cannot handle code ending with multiline statements
     check_size(example, 3)
 
 
+# We currently cannot handle code ending with comment lines.
 def test_size_comment_after_last_line():
     example = """
 class Foo:
@@ -196,7 +198,6 @@ class Foo:
     check_size(example, 4)
 
 
-@skip_if_not_has_async
 def test_size_generator():
     example = """
 class Foo:
@@ -207,26 +208,13 @@ class Foo:
 
 
 @skip_if_not_has_async
-def test_size_coroutines_basic():
+def test_size_async_function_def():
     example = """
 class Foo:
     async def foo(some_attr):
         pass
 """
     check_size(example, 3)
-
-
-@skip_if_not_has_async
-def test_size_await():
-    example = """
-class Foo:
-    async def __aenter__(self):
-        await log('entering context')
-
-    async def __aexit__(self, exc_type, exc, tb):
-        await log('exiting context')
-"""
-    check_size(example, 6)
 
 
 @skip_if_not_has_async
@@ -245,6 +233,7 @@ def test_size_async_for():
     example = """
 class Foo:
     async def foo():
-        async for a in b: pass
+        async for a in b:
+            pass
 """
-    check_size(example, 3)
+    check_size(example, 4)
--- a/tests/test_size.py
+++ b/tests/test_size.py
@@ -1,12 +1,20 @@
 import ast
 
-from vulture.lines import estimate_lines
+from vulture.lines import estimate_lines, count_lines
 
 
-def check_size(example, span):
+def check_size(example, size):
     # offset of one because 'node' is ast.Module object
     node = ast.parse(example)
-    assert estimate_lines(node) - 1 == span
+    assert estimate_lines(node) - 1 == size
+
+
+def check_span(example, span):
+    tree = ast.parse(example)
+    results = []
+    for node in tree.body:
+        results.append(count_lines(node))
+    assert span == results
 
 
 def test_size_function():
@@ -18,6 +26,7 @@ def func():
     return len(sys.argv)
 """
     check_size(example, 5)
+    check_span(example, [5])
 
 
 def test_size_class():
@@ -34,6 +43,7 @@ class Foo(object):
         return len(sys.argv)
 """
     check_size(example, 9)
+    check_span(example, [10])
 
 # TODO improve estimate_lines to count the "else" clauses
 # and the estimate will get better (higher).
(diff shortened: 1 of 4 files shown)


jendrikseipp:
As long as all changes are related to the PR's main goal, it's fine.

vulture/lines.py
```diff
@@ -1,12 +1,14 @@
 import ast
-from functools import wraps
+from collections import namedtuple
 import sys
 
+ENTRY_POINTS = (ast.ClassDef, ast.FunctionDef)
```
jendrikseipp: These entry points were only for testing. The code should work for all types of nodes.

vulture/lines.py
```diff
@@ -23,27 +25,26 @@
...
-        try:
-            return _cache[args]
-        except KeyError:
-            _cache[args] = result = func(*args)
-            return result
-    return wrapped
+        last_line = last_body.lineno
+        SourceRange = namedtuple("SourceRange",
+                                 ['name', 'start_line', 'last_line'])
+        source_range = SourceRange(entry.name, entry.lineno, last_line)
```
jendrikseipp: This is an odd design: the function returns entry.name and entry.lineno, but this info is already known to the caller of the function.

vulture/lines.py
```diff
@@ -23,27 +25,26 @@
...
+        yield source_range
```
jendrikseipp: Functions that only return one thing, shouldn't be generators. I think the function should just be called `get_last_line_number(node)` and `return last_line`.

vulture/lines.py
```diff
@@ -23,27 +25,26 @@
     })
 
 
-# Reinventing wheel here to keep vulture lightweight.
-def memoize(func):
-    _cache = {}
+def _get_range(entry):
```
jendrikseipp: entry -> node

vulture/lines.py
```diff
@@ -23,27 +25,26 @@
...
-    sorting the list of unused code that a developer might want to remove.
-    """
-    return 1 + sum(estimate_lines(child)
-                   for field in TRAVERSABLE_FIELDS.get(node.__class__, ())
-                   for child in getattr(node, field))
+def count_lines(code):
+    span = 1
+    for source_range in _get_range(code):
+        span = (1 + source_range.last_line - source_range.start_line)
+    return span
```
jendrikseipp: I think this can just be 
    
    def count_lines(node):
        return get_last_line_number(node) - node.lineno + 1

vulture/lines.py
```diff
@@ -23,27 +25,26 @@
...
+def _get_range(entry):
+    if isinstance(entry, ENTRY_POINTS):
+        last_body = entry.body[-1]
+        while isinstance(last_body, tuple(TRAVERSABLE_FIELDS.keys())):
+            fields = TRAVERSABLE_FIELDS[last_body.__class__]
+            for field in reversed(fields):
+                children = getattr(last_body, field)
+                if children:
+                    break
+            last_body = children[-1]  # The last child.
```
jendrikseipp: Probably clearer:

    if children:
        last_body = children[-1]
        break

The comment is not needed.

tests/test_size.py
```diff
@@ -1,12 +1,14 @@
 import ast
 
-from vulture.lines import estimate_lines
+from vulture.lines import count_lines
 
 
 def check_size(example, span):
```
jendrikseipp: span -> size

jendrikseipp: The second argument should be a single number, not a list:

    def check_size(code, size):
        module = ast.parse(code)
        for node in module.body:
            if isinstance(node, ast.ClassDef) and node.name == 'Foo':
                assert count_lines(node) == size
        assert False, 'failed to find top-level class "Foo" in code'

vulture/lines.py
RJ722: Previously, I was passing all this around for tackling multi-line return statement. But then, I thought it would be better if we could get this up and running. 
It is indeed odd. 😛

tests/test_size.py
```diff
@@ -1,12 +1,14 @@
...
-    node = ast.parse(example)
-    assert estimate_lines(node) - 1 == span
+    tree = ast.parse(example)
+    results = []
+    for node in tree.body:
+        results.append(count_lines(node))
+    assert span == results
 
 
 def test_size_function():
```
jendrikseipp: I think this test case can be removed.

tests/test_size.py
```diff
@@ -56,56 +58,93 @@ def bar(self):
...
-else:
-    pass
+class Foo:
+    while "b" > "a":
+        pass
+    else:
+        pass
 """
-    check_size(example, 7)
+    check_size(example, [5])
```
jendrikseipp: we also need to test while without else

tests/test_size.py
```diff
@@ -56,56 +58,93 @@ def bar(self):
...
+        x = sys.argv[99]
+    except IndexError:
+        pass
+    except Exception:
+        pass
+    else:
+        pass
 """
-    check_size(example, 3)
+    check_size(example, [9])
```
jendrikseipp: please also check try without else with **two** excepts.

tests/test_size.py
```diff
@@ -56,56 +58,93 @@ def bar(self):
...
-    print("loop")
-else:
-    print("else")
+class Foo:
+    for arg in sys.argv:
+        print("loop")
+    else:
+        print("else")
+"""
+    check_size(example, [5])
```
jendrikseipp: please also check `for` without `else`

tests/test_size.py
```diff
@@ -56,56 +58,93 @@ def bar(self):
...
+
+
+def test_size_class_nested():
+    example = """
+class Foo:
+    class test:
```
jendrikseipp: test -> Bar

tests/test_size.py
```diff
@@ -56,56 +58,93 @@ def bar(self):
...
+        pass
+"""
+    check_size(example, [3])
+
+
+def test_multi_line_return():
+    example = """
+def long_string_return(o):
```
jendrikseipp: This can become

    class Foo:
        def foo():
            return (
                'very'
                'long'
                'string')

Please add a comment: `// We currently cannot handle code ending with multiline statements.`

tests/test_size.py
```diff
@@ -56,56 +58,93 @@ def bar(self):
...
+    return '''I am a
+    veryyy
+    long
+    string.
+    '''
 """
-    check_size(example, 3)
+    check_size(example, [2])
```
jendrikseipp: We also need a test for 

    class Foo:
        foo = 1
        bar = 2

tests/test_size.py
```diff
@@ -56,56 +55,125 @@ def bar(self):
...
+
+
+def test_multi_line_return():
+    example = """
+class Foo:
+    def long_string_return(o):
+        return (
+            'very'
+            'long'
+            'string')
```
RJ722: I wonder what would happen if we add a comment after last_line, 😕 !

RJ722: Not detected, :-(

vulture/lines.py
```diff
@@ -23,27 +22,18 @@
     })
 
 
-# Reinventing wheel here to keep vulture lightweight.
-def memoize(func):
-    _cache = {}
+def get_last_line_number(node):
+    last_body = node.body[-1]
```
jendrikseipp: The fact that we use [-1] twice suggests that this can be simplified. I think we should remove the line `last_body = node.body[-1]` and rename all occurences of "last_body" to "node" in the code below. This will probably make some tests fail. However, the current code doesn't handle all cases correctly either. For example, it fails if node doesn't have a field "body". I think the following code might be closer to what we're after (untested):

```
def _get_last_child(node):
    for name in ['finalbody', 'orelse', 'handlers', 'returns', 'body']:
        try:
            field = getattr(node, name)
        except AttributeError:
            continue

        if isinstance(field, list):
            if field:
                return field[-1]
        else:
            return field
    return None


def get_last_line_number(node):
    while True:
        last_child = _get_last_child(node)
        if last_child is None:
            return node.lineno
        node = last_child
```

RJ722: The design is great, it even determines the last line for multi-line return statements and comments after return correctly (for python3). 😄

vulture/lines.py
```diff
@@ -1,12 +1,11 @@
...
 TRAVERSABLE_FIELDS = {
-        ast.ClassDef: ('body', 'decorator_list'),
+        ast.ClassDef: ('decorator_list', 'body', ),
         ast.ExceptHandler: ('body',),
         ast.For: ('body', 'orelse'),
-        ast.FunctionDef: ('body', 'decorator_list'),
+        ast.FunctionDef: ('decorator_list', 'body',),
         ast.If: ('body', 'orelse'),
         ast.Module: ('body',),
         ast.While: ('body', 'orelse'),
```
RJ722: Since `ast.FunctionDef.returns` returns `None`, it messes up with the 'fields'. We probably won't need it.

tests/test_size.py
```diff
@@ -56,56 +59,190 @@ def bar(self):
...
+    example = """
+class Foo:
+    async def bar():
+        async with x:
+            pass
+"""
+    check_size(example, 4)
+
+
+@pytest.mark.skipif(not PY35, reason='requires Python3.5')
```
jendrikseipp: Can you define a decorator `skip_if_not_has_async` once to avoid repeating the code?

vulture/lines.py
```diff
@@ -1,49 +1,26 @@
...
-    TRAVERSABLE_FIELDS.update({
-        ast.TryExcept: ('body', 'handlers', 'orelse'),
-        ast.TryFinally: ('body', 'finalbody'),
-    })
-else:
-    TRAVERSABLE_FIELDS.update({
-        ast.Try: ('body', 'handlers', 'orelse', 'finalbody')
-    })
+def _get_last_child(node):
+    fields = ['finalbody', 'orelse', 'handlers', 'body']
```
jendrikseipp: Also, please rename `fields` to `reverse_ordered_fields`.

tests/test_size.py
```diff
@@ -33,17 +39,14 @@ def func():
         import sys
         return len(sys.argv)
 """
-    check_size(example, 9)
+    check_size(example, 10)
 
 # TODO improve estimate_lines to count the "else" clauses
 # and the estimate will get better (higher).
```
jendrikseipp: Please remove outdated comment.

RJ722:
> Why did you remove 'returns'?

Since ast.FunctionDef.returns` returns `None`, it messes. up with the algorithm. We probably won't need it.

tests/test_size.py
```diff
@@ -1,23 +1,44 @@
...
 
-def check_size(example, span):
-    # offset of one because 'node' is ast.Module object
-    node = ast.parse(example)
-    assert estimate_lines(node) - 1 == span
+HAS_ASYNC = getattr(ast, 'AsyncFunctionDef', None)
 
 
-def test_size_function():
+def skip_if_not_has_async(item):
```
jendrikseipp: Could you please give me write access to your branch? I have some minor nitpicks that I'd like to quickly change.

RJ722: I have enabled "Allow edits from maintainers". Do I need to do something else?

jendrikseipp: Thanks. It seems to work fine, I just pushed a test commit.

jendrikseipp:
I'm done with my changes. Let's see if the "wild" code turns up anything :-)

RJ722:
Should we add a test for closures as well?

jendrikseipp:
I don't think we need tests for closures.

Issue #67, opened 2017-07-09
Let Item inherit from 'object'


RJ722:
Let Item inherit from "object", not from "str". The new class needs the members "name", ```__eq__``` (returns name) and ```__hash__``` (returns hash(name)).

PR #71, opened 2017-07-18
Let Item inherit from object

RJ722:
Closes: #67

### DIFF
--- a/tests/test_format_strings.py
+++ b/tests/test_format_strings.py
@@ -1,6 +1,7 @@
 import pytest
-
 from vulture import Vulture
+
+from . import check
 
 
 @pytest.fixture
@@ -10,15 +11,15 @@ def v():
 
 def test_old_format_string(v):
     v.scan("a = 1\n'%(a)s, %(b)d' % locals()")
-    assert v.defined_funcs == []
-    assert v.defined_vars == ['a']
-    assert v.used_vars == ['a', 'b', 'locals']
-    assert v.unused_vars == []
+    check(v.defined_funcs, [])
+    check(v.defined_vars, ['a'])
+    check(v.used_vars, ['a', 'b', 'locals'])
+    check(v.unused_vars, [])
 
 
 def test_new_format_string(v):
     v.scan("a = 1\n'{a}, {b}'.format(**locals())")
-    assert v.defined_funcs == []
-    assert v.defined_vars == ['a']
-    assert v.used_vars == ['a', 'b', 'locals']
-    assert v.unused_vars == []
+    check(v.defined_funcs, [])
+    check(v.defined_vars, ['a'])
+    check(v.used_vars, ['a', 'b', 'locals'])
+    check(v.unused_vars, [])
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -55,8 +55,12 @@ if sys.version_info < (3, 4):
 IGNORED_IMPORTS = ["*"]
 
 
-def _get_unused_items(defined, used):
-    return list(sorted(set(defined) - set(used), key=lambda x: x.lower()))
+def _get_unused_items(defined_items, used_names):
+    used_names = set(used_names)
+    unused_items = [item for item in set(defined_items)
+                    if item.name not in used_names]
+    unused_items.sort(key=lambda item: item.name.lower())
+    return unused_items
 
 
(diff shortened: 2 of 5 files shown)


vulture/core.py
```diff
@@ -56,21 +56,29 @@
...
+        self.typ = typ
+        self.filename = filename
+        self.lineno = lineno
+        self.size = size
+
+    def __str__(self):
+        return self.name
+
+    def __eq__(self, other):
+        return self.name
```
jendrikseipp: should be

    return self.name == self.other

vulture/core.py
```diff
@@ -91,14 +91,22 @@ def _ignore_variable(filename, varname):
...
+        self.typ = typ
+        self.filename = filename
+        self.lineno = lineno
+        self.size = size
+
+    def __str__(self):
+        return self.name
+
+    def __eq__(self, other):
+        return self.name == other
```
jendrikseipp: other -> other.name

vulture/core.py
```diff
@@ -185,7 +193,7 @@ def exclude(name):
                 self.scan(module_string, filename=module)
 
             for name in self.defined_imports:
```
jendrikseipp: name -> defined_import

vulture/core.py
RJ722: That would require us to compare it with only `Item` objects.
 IMHO, this would cater to a better UX.
For ex, when making use of the API, it would be better if we could just compare it with the "name" string.

jendrikseipp: I don't think so. The current code will look like a bug to many people. If we want to compare Item objects to strings, we need to do `item.name == mystring`.

jendrikseipp: You'll probably have to change the set difference operation in _get_unused_items() since some lists contain Items and others contain strings. I just pushed a commit which unifies things: All defined_* lists always contained Item objects, all remaining lists ("usages") now contain plain strings. What we probably need is

```
def _get_unused_items(defined_items, used_names):
    used_names = set(used_names)
    unused_items = [item for item in set(defined_items) if item.name not in used_names]
    unused_items.sort(key=lambda item: item.name.lower())
    return unused_items
```

RJ722: @jendrikseipp  Thank you for the help, I was stuck on that one.

jendrikseipp: Sure, next time just ask when you're stuck :-)

tests/test_utils.py
```diff
@@ -0,0 +1,3 @@
+def check_item(asserted, assertion):
+    names = set(str(getattr(item, 'name', item)) for item in asserted)
```
jendrikseipp: add `# TODO: After converting used_vars and used_attrs to sets, use dedicated tests for them.`

tests/test_format_strings.py
```diff
@@ -1,5 +1,6 @@
 import pytest
 
+from . import check
 from vulture import Vulture
```
jendrikseipp:
```
import pytest
from vulture import Vulture

from . import check
```
We should treat vulture as any other non-stdlib module. (This applies to the other test files as well)

tests/test_utils.py
RJ722: I made a slight change, as it wasn't fitting in a single line. Is it okay?

vulture/core.py
```diff
@@ -91,14 +95,22 @@ def _ignore_variable(filename, varname):
...
+class Item(object):
+    def __init__(self, name, typ, filename, lineno, size=1):
+        self.name = name
+        self.typ = typ
+        self.filename = filename
+        self.lineno = lineno
+        self.size = size
+
+    def __str__(self):
+        return self.name
```
jendrikseipp: Can we remove the `__str__` function and use item.name explicitly, please?

tests/test_utils.py
jendrikseipp: Sure :-)

tests/test_format_strings.py
RJ722: Should I change core.py as well?

jendrikseipp: It doesn't need changing since all other imported modules are from the stdlib.

jendrikseipp:
Thanks!

PR #75, opened 2017-07-22
Whitelist unittest

RJ722:
Create whitelist file for the module - unittest

## Checklist:

### DIFF
--- a/vulture/whitelists/unittest.py
+++ b/vulture/whitelists/unittest.py
@@ -1,0 +1,11 @@
+from whitelist_utils import Whitelist
+
+whitelist_test_case = Whitelist()
+whitelist_test_case.setUp
+whitelist_test_case.tearDown
+whitelist_test_case.setUpClass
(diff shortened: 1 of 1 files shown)


vulture/whitelists/unittest.py
```diff
@@ -0,0 +1,7 @@
+from whitelist_utils import Whitelist
+
+test_case = Whitelist()
```
jendrikseipp: test_case -> whitelist_test_case

We have to make sure we don't hide a different variable named `test_case`.

vulture/whitelists/unittest.py
```diff
@@ -0,0 +1,7 @@
...
+test_case.setUp
+test_case.tearDown
+test_case.setUpClass
+test_case.tearDownClass
```
jendrikseipp: The following methods are missing: run, skipTest, subTest, debug

PR #77, opened 2017-07-24
Exclude 'settings.py' and 'conf.py' by default

RJ722:
Since, neither `settings.py` nor `conf.py` contain any "logical" parts of application, they can be excluded by default.


jendrikseipp:
I think this change is too drastic. Some projects may have conf.py files that should be checked. We need to figure out a way to load the list of ignored files based on the imports.

jendrikseipp:
I thought about this some more and think the best solution for cases like these is to use the `--exclude` option.

Issue #80, opened 2017-07-25
Travis CI failing


RJ722:
IMHO, it's because Travis recently changed the default environment from `precise` to `trusty` and this had side effects which in turn cause some errors to be raised when testing for`pypi`, therefore failing CI build every time. 

I think, an explicit mention of `dist: precise` in `.travis.yml` will fix it.

PR #79, opened 2017-07-25
Distribute vulture as a wheel

RJ722:
Closes: #78

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation in the README file.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #81, opened 2017-07-25
Fix Travis CI 

RJ722:
Closes: #80

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
 - '3.5'
 - '3.6'
 - 3.7-dev
-- pypy2-5.8.0
+- pypy2.7-5.8.0
 
 matrix:
   fast_finish: true


jendrikseipp:
I think we should be using the Trusty images. I tried to apply a different fix from https://github.com/travis-ci/travis-ci/issues/6865 and https://github.com/travis-ci/travis-cookbooks/pull/861, but so far no luck... It would be great if you could investigate some more. If you don't find a way to use pypy on Trusty, we can use the Precise images.

RJ722:
Hi @jendrikseipp 
I just realised I have named the PR and the commit wrongly. Changing them.

Do you think it is a good idea to test these on the master branch?

RJ722:
Strange, if we change `pypy2-5.8.0` to `pypy2.7-5.8.0`, it works 😅 

Looks like a upstream bug?

RJ722:
Although https://github.com/travis-ci/travis-ci/issues/6865#issuecomment-317819856 mentions that it should now work for every infrastructure, it doesn't works for commits prior to this one. I tried restarting the build for an other PR (#79), but still no effect.  :-(

jendrikseipp:
You're right, I shouldn't have tried to fix this in master. Thanks for looking into this!

Issue #82, opened 2017-07-27
Report any code after return statements


RJ722:
Report any code which is written after 'return' statement in the same logic level in a function.
For example:
```python
def foo():
    return bar()
    print("Hello")
```

This should report Line 3 as dead code.

PR #83, opened 2017-07-27
Report code after return statements

RJ722:
## Description

Check and report any nodes which appear after `ast.Return` while visiting function definition.

## Related Issue

Closes: #82

## Checklist:

- [X] All ~~new and~~ existing tests passed.

### DIFF
--- a/tests/test_unreachable.py
+++ b/tests/test_unreachable.py
@@ -1,9 +1,13 @@
 from . import v
+assert v  # Silence pyflakes
 
 
-def check_unreachable_lines(assertion, asserted):
-    unreachable_lines = [node.lineno for node in assertion]
-    assert unreachable_lines == asserted
+def check_unreachable(v, lineno, size, name):
+    assert len(v.unreachable_code) == 1
+    item = v.unreachable_code[0]
+    assert item.lineno == lineno
+    assert item.size == size
+    assert item.name == name
 
 
 def test_assignment(v):
@@ -13,14 +17,14 @@ def foo():
     return
     a = 1
 """)
-    check_unreachable_lines(v.unreachable_code, [4])
+    check_unreachable(v, 4, 1, 'return')
 
 
 def test_multiline_return_statements(v):
     v.scan("""\
 def foo():
     print("Something")
-    return (Something,
+    return (something,
             that,
             spans,
             over,
@@ -28,7 +32,7 @@ def foo():
             lines)
     print("Hello World")
 """)
-    check_unreachable_lines(v.unreachable_code, [9])
+    check_unreachable(v, 9, 1, 'return')
 
 
 def test_multiple_return_statements(v):
@@ -36,11 +40,9 @@ def test_multiple_return_statements(v):
 def foo():
     return something
     return None
-    return (Some,
-           big,
-           statemnt)
+    return (some, statement)
 """)
-    check_unreachable_lines(v.unreachable_code, [3, 4])
+    check_unreachable(v, 3, 2, 'return')
 
 
 def test_pass(v):
@@ -48,18 +50,57 @@ def test_pass(v):
 def foo():
     return
     pass
-    return Something
+    return something
 """)
-    check_unreachable_lines(v.unreachable_code, [3])
+    check_unreachable(v, 3, 2, 'return')
+
+
+def test_multiline_return(v):
+    v.scan("""
+def foo():
+    return \
+        "Hello"
+    print("Unreachable code")
+""")
+    check_unreachable(v, 4, 1, 'return')
 
 
 def test_recursive_functions(v):
     v.scan("""\
 def foo(a):
-    if a == 1:
+    if a==1:
         return 1
     else:
         return foo(a-1)
         print("This line is never executed")
 """)
-    check_unreachable_lines(v.unreachable_code, [6])
+    check_unreachable(v, 6, 1, 'return')
+
+
+def test_semicolon(v):
+    v.scan("""\
+def foo():
+    return; a = 1
+""")
+    check_unreachable(v, 2, 1, 'return')
+
+
+def test_list(v):
+    v.scan("""\
+def foo(a):
+    return
+    a[1:2]
+""")
+    check_unreachable(v, 3, 1, 'return')
+
+
+def test_continue(v):
+    v.scan("""\
+def foo():
+    if foo():
+        return True
+        continue
+    else:
+        return False
+""")
+    check_unreachable(v, 4, 1, 'return')
--- a/tests/test_unreachable.py
+++ b/tests/test_unreachable.py
@@ -1,0 +1,65 @@
+from . import v
+
+
+def check_unreachable_lines(assertion, asserted):
+    unreachable_lines = [node.lineno for node in assertion]
+    assert unreachable_lines == asserted
+
+
+def test_assignment(v):
+    v.scan("""\
+def foo():
+    print("Hello World")
+    return
+    a = 1
+""")
+    check_unreachable_lines(v.unreachable_code, [4])
+
+
+def test_multiline_return_statements(v):
+    v.scan("""\
+def foo():
+    print("Something")
+    return (Something,
+            that,
+            spans,
+            over,
+            multiple,
+            lines)
+    print("Hello World")
+""")
+    check_unreachable_lines(v.unreachable_code, [9])
+
+
+def test_multiple_return_statements(v):
+    v.scan("""\
+def foo():
+    return something
+    return None
+    return (Some,
+           big,
+           statemnt)
+""")
+    check_unreachable_lines(v.unreachable_code, [3, 4])
+
+
+def test_pass(v):
+    v.scan("""\
+def foo():
+    return
+    pass
+    return Something
+""")
+    check_unreachable_lines(v.unreachable_code, [3])
+
+
+def test_recursive_functions(v):
+    v.scan("""\
+def foo(a):
+    if a == 1:
+        return 1
+    else:
+        return foo(a-1)
+        print("This line is never executed")
+""")
+    check_unreachable_lines(v.unreachable_code, [6])
--- a/tests/test_unreachable.py
+++ b/tests/test_unreachable.py
@@ -68,10 +68,10 @@ def foo():
 def test_recursive_functions(v):
     v.scan("""\
 def foo(a):
-    if a==1:
+    if a == 1:
         return 1
     else:
-        return foo(a-1)
+        return foo(a - 1)
         print("This line is never executed")
 """)
     check_unreachable(v, 6, 1, 'return')
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -104,7 +104,7 @@ class Item(object):
         self.filename = filename
         self.lineno = lineno
         self.size = size
-        self.message = message or "Unused {typ} '{name}'".format(**locals())
+        self.message = message or "unused {typ} '{name}'".format(**locals())
 
     def _tuple(self):
         return (self.filename, self.lineno, self.name)
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -104,8 +104,7 @@ class Item(object):
         self.filename = filename
         self.lineno = lineno
         self.size = size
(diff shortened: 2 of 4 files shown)


jendrikseipp:
I think what we need is to loop over all **lists** of AST nodes in the given tree (see Python/Lib/ast.py for how to do something similar, see functions walk() and iter_child_nodes()).

Then for each list of AST nodes, we need to check whether there is a Return node that's not at the end of the list. If it's not at the end, we return an Item that start at the node after the Return node.

Also, we need tests :-) The new method can go into vulture/unreachable.py and the tests into test_unreachable.py.

vulture/utils.py
```diff
@@ -37,6 +38,17 @@ def read_file(filename):
...
+def return_nodes_after_return(tree):
+    try:
+        for i, node in enumerate(tree.body, 0):
+            if isinstance(node, ast.Return):
+                return tree.body[i+1:]
+    except AttributeError:
+        return []
+    else:
+        return return_nodes_after_return(node)
+
```
RJ722: I tried some basic implementations using `walk` and `iter_child_nodes`. Like:
```python
def return_node_after_return(tree):
    """
    Iterate over list of every node present in each level of the ast.
    Return the item just after return statement.
    """
    for parent_node in ast.walk(tree):
        child_nodes = [child_node for child_node in ast.iter_child_nodes(parent_node)]
        for i, node in enumerate(child_nodes):
            if isinstance(node, ast.Return):
                try:
                    return child_nodes[i+1]
                except IndexError:
                    return None
```
But, lies a pitfall: The _if-else_, _try-except_ clauses are a single node. Thus it would be more difficult to list items in the 'same logical level'. This is also sub-optimal performance wise, but I guess, we can improve it.

Is the current approach not feasible?

jendrikseipp: I don't think either solution is generic enough. I think we need something like this:

```python
def get_ast_lists(tree):
    for node in ast.walk(tree):
        for name, field in ast.iter_fields(node):
            if isinstance(field, list):
                yield field


def search_unreachable_code(tree):
    """
    Iterate over all lists of AST nodes in the given tree. Then for each
    list, check whether there is a Return node that's not at the end of
    the list. If it's not at the end, return an Item that starts at the
    node after the Return node.

    """
    for ast_list in get_ast_lists(tree):
        for index, node in enumerate(ast_list):
            if isinstance(node, ast.Return):
                try:
                    unreachable_node = ast_list[index + 1]
                except IndexError:
                    pass
                else:
                    yield Item(
                        'code after return', 'unreachable code',
                        'filename=todo', unreachable_node.lineno)
```

jendrikseipp: I think it's best to incorporate this functionality into the current design by adding this method at the end of the Vulture methods

```python
    def generic_visit(self, node):
        """Called if no explicit visitor function exists for a node."""
        for field, value in ast.iter_fields(node):
            if isinstance(value, list):
                self._handle_ast_list(value)
                for item in value:
                    if isinstance(item, ast.AST):
                        self.visit(item)
            elif isinstance(value, ast.AST):
                self.visit(value)
```
and this method above the `visit()` method:

```python
    def _handle_ast_list(self, ast_list):
        for index, node in enumerate(ast_list):
            if isinstance(node, ast.Return):
                try:
                    first_unreachable_node = ast_list[index + 1]
                except IndexError:
                    pass
                else:
                    self._define(
                        self.unreachable_code,
                        'code after return',
                        first_unreachable_node.lineno,
                        size=(
                            lines._get_last_line_number(ast_list[-1]) -
                            first_unreachable_node.lineno + 1))
```

The most important part, however, is creating test cases. Remember that return statements can occur in very many places in Python.

vulture/core.py
```diff
@@ -231,25 +232,29 @@ def by_size(item):
...
             else:
                 size_report = ''
-
-            print("{0}:{1:d}: Unused {2} '{3}'{4}".format(
+            name_report = ("'{0}'".format(item.name) if item.name != ''
+                           else "")
+            print("{0}:{1:d}: Unused {2} {3}{4}".format(
                 utils.format_path(item.filename), item.lineno, item.typ,
-                item.name, size_report))
+                name_report, size_report))
```
RJ722: I was quite unclear as to what should be printed when reporting unreachable code, so decided to print nothing instead. Does it seems sound?

jendrikseipp: The reports should look like

```
bar.py:10: unused function 'myfunc'
foo.py:12: unreachable code after return statement
foobar.py14: unreachable code in body of while loop  # We'll implement this later.
```

To achieve that, let's add a field `Item.message` to the Item class (and a `message=''` kwarg to it's constructor and the `_define` function). If `not message` in the Item constructor, we set `Item.message` to `"unused {typ} '{name}'".format(**locals())`.

The new unused-code-after-return Items objects can get a `name='return'`.

vulture/core.py
```diff
@@ -415,6 +420,33 @@ def visit(self, node):
             visitor(node)
         return self.generic_visit(node)
 
+    def _handle_ast_list(self, ast_list, parent_node):
```
jendrikseipp: needs a docstring.

jendrikseipp: parent_node is unneeded, I think.

vulture/core.py
```diff
@@ -415,6 +420,33 @@ def visit(self, node):
...
+        for index, node in enumerate(ast_list):
+            if isinstance(node, ast.Return):
+                try:
+                    first_unreachable_node = ast_list[index + 1]
+                except IndexError:
+                    pass
```
jendrikseipp: let's `continue` here and remove the `else` clause. Then the code can be moved to the left by four spaces.

vulture/core.py
RJ722: Oh! yes, I was experimenting with some thing, and this change also got committed accidentally.

tests/test_unreachable.py
```diff
@@ -0,0 +1,65 @@
...
+def test_recursive_functions(v):
+    v.scan("""\
+def foo(a):
+    if a == 1:
+        return 1
+    else:
+        return foo(a-1)
+        print("This line is never executed")
+""")
+    check_unreachable_lines(v.unreachable_code, [6])
```
RJ722: IMHO, we'll be needing more tests for:
* nested if/else clauses having return statements
* nested try/except statements
* for/else, for, while/else and while loops

Did I leave something?

tests/test_unreachable.py
```diff
@@ -0,0 +1,65 @@
+from . import v
+
+
+def check_unreachable_lines(assertion, asserted):
+    unreachable_lines = [node.lineno for node in assertion]
+    assert unreachable_lines == asserted
```
jendrikseipp: I propose

```
def check_unreachable(v, line, size, name):
    assert len(v.unreachable_code) == 1
    item = v.unreachable_code[0]
    assert item.line == line
    assert item.size == size
    assert item.name == name
```

tests/test_unreachable.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_multiple_return_statements(v):
+    v.scan("""\
+def foo():
+    return something
+    return None
+    return (Some,
+           big,
+           statemnt)
```
jendrikseipp: typo, and can be single line.

tests/test_unreachable.py
```diff
@@ -0,0 +1,65 @@
...
+""")
+    check_unreachable_lines(v.unreachable_code, [3, 4])
+
+
+def test_pass(v):
+    v.scan("""\
+def foo():
+    return
+    pass
+    return Something
```
jendrikseipp: lowercase

tests/test_unreachable.py
```diff
@@ -0,0 +1,65 @@
...
+    check_unreachable_lines(v.unreachable_code, [3])
+
+
+def test_recursive_functions(v):
+    v.scan("""\
+def foo(a):
+    if a == 1:
+        return 1
+    else:
+        return foo(a-1)
```
jendrikseipp: `a - 1`` (spaces)

tests/test_unreachable.py
jendrikseipp: I think we're already on the same page, but just to be clear: we only want to catch unreachable code after return statements in this PR, we'll cover the other cases later.

For tests, we need to think about edge cases and stuff that can go wrong with the current code. Multiple return statements is a good idea for a test case and it uncovered a design error already (see my `return` comment on the code). What other edge cases are there?

```
def foo():
    return; a = 1
```
```
def foo():
    return
    a[1:2]
```

```
class Foo():
    return
    a = 1
```

I don't think we need the additional tests you mention in your message.

vulture/core.py
```diff
@@ -415,6 +420,37 @@ def visit(self, node):
...
+                try:
+                    first_unreachable_node = ast_list[index + 1]
+                except IndexError:
+                    continue
+                self._define(
+                    self.unreachable_code,
+                    'return',
+                    first_unreachable_node.lineno,
+                    size=(
+                        lines._get_last_line_number(ast_list[-1]) -
```
RJ722: Oh! okay. Rebasing the branch now.

vulture/core.py
```diff
@@ -232,9 +235,10 @@ def by_name(item):
             return (item.filename.lower(), item.lineno)
 
         return sorted(
-            self.unused_attrs + self.unused_classes + self.unused_funcs +
-            self.unused_imports + self.unused_props + self.unused_vars,
-            key=by_size if self.sort_by_size else by_name)
+                self.unused_attrs + self.unused_classes + self.unused_funcs +
+                self.unused_imports + self.unused_props + self.unused_vars +
+                self.unreachable_code, key=by_size if self.sort_by_size
+                else by_name)
```
RJ722: Sorry, I didn't notice this until now.

tests/test_unreachable.py
RJ722: > 
```python
class Foo():
    return
    a = 1
```
Won't this throw up `SyntaxError: 'return' outside function`?

RJ722: Also, I just wanted to check for "for/while loops" and other things that span over multiple lines and occur after return statements. Are you sure, we don't need tests for them?

jendrikseipp: That's right, I forgot that we can only use `return` in functions.

I don't think we need the other tests. The code doesn't care about the structure of what comes after the return statement.

vulture/core.py
RJ722: I misunderstood and thought that this change has been done upstream and I need to pull it in. :p 

Done!

tests/test_unreachable.py
```diff
@@ -0,0 +1,84 @@
...
+    check_unreachable(v, 3, 2, 'return')
+
+
+def test_recursive_functions(v):
+    v.scan("""\
+def foo(a):
+    if a==1:
+        return 1
+    else:
+        return foo(a-1)
```
jendrikseipp: even though this is text, it should follow our coding conventions: `a == 1` and `a - 1`.

vulture/core.py
```diff
@@ -98,12 +98,13 @@ class Item(object):
...
     Hold the name, type and location of defined code.
     """
-    def __init__(self, name, typ, filename, lineno, size=1):
+    def __init__(self, name, typ, filename, lineno, size=1, message=''):
         self.name = name
         self.typ = typ
         self.filename = filename
         self.lineno = lineno
         self.size = size
+        self.message = message or "Unused {typ} '{name}'".format(**locals())
```
jendrikseipp: unused (lowercase)

RJ722: Done, Thanks!

tests/test_unreachable.py
RJ722: Again, misread. So sorry, not one of my best days.

RJ722:
Thanks for the help @jendrikseipp :-)

jendrikseipp:
Thank you!

jendrikseipp:
I think the next item would be to detect unreachable code after ``break``, ``continue`` and ``raise`` statements. The main part here is adding tests, I guess.

Issue #88, opened 2017-08-10
detect unreachable code if the condition is unsatisfiable


RJ722:
As, the TODO file says, it would consists of:

- [X] Detect unreachable code à la while False: (by implementing Vulture.visit_While(self, node).
- [ ] Detect unreachable code à la if False: (by implementing Vulture.visit_If(self, node).
- [X] ~~Detect unreachable else clauses by detecting if True:.~~ (#90)
- [ ] Detect unreachable code for ast.IfExp.
- [X] ~~Detect unreachable code for ast.Assert.~~

RJ722:
> Detect unreachable code for ast.Assert.

Won't assert statements throw up errors if unsatisfied?

jendrikseipp:
`assert False` means that the code is either never reached or will throw an AssertionError when reached. But you're probably right. There may be cases where `assert False` is intended. Let's not try to detect unsatisfiable assert statements.

RJ722:
Hi @jendrikseipp, can we reopen the issue because `if False` statements are yet not reported.

> Detect unreachable code for `ast.IfExp`.

Is this any different from `ast.If`?

jendrikseipp:
Let's do ast.If before we look into ast.IfExp. We can handle ast.IfExp in a separate PR. An example for ast.IfExp is `a = 1 if False else 2`.

jendrikseipp:
I don't think it's worth the effort to catch unsatisfiable ast.IfExp conditions. Therefore, I think we're done here :-)

PR #89, opened 2017-08-11
Catch unsatisfiable while condition

RJ722:
Closes: #88 

## Checklist:

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -96,6 +96,20 @@ def _ignore_variable(filename, varname):
         _is_special_name(varname))
 
 
+def condition_is_unsatisfiable(condition):
+    """
+    Checks if the given condition is unsatisfiable 
+    """
+    try:
+        cond = condition if isinstance(condition, ast.Expression) else ast.Expression(condition)
+        test = compile(cond, filename="<string>", mode='eval')
+        satisfiable = bool(eval(test, {}, {}))
+    except (NameError, ValueError):
+        return False
+    else:
+        return not satisfiable
+
+
 class Item(object):
     """
     Hold the name, type and location of defined code.
@@ -378,6 +392,15 @@ class Vulture(ast.NodeVisitor):
             if param and isinstance(param, str):
                 self._define_variable(param, node.lineno, confidence=100)
 
+
+    def visit_If(self, node):
+        if condition_is_unsatisfiable(node.test):
+            self._define(self.unreachable_code, 'if', node.lineno,
+                size=lines.count_lines(node),
+                message="unsatisfiable 'if' condition",
+                confidence=DEFAULT_CONFIDENCE)
+
+
     def visit_Import(self, node):
         self._add_aliases(node)
 
@@ -424,6 +447,15 @@ class Vulture(ast.NodeVisitor):
             for attr in name_and_attrs[1:]:
                 if is_identifier(attr):
                     self.used_attrs.add(attr)
+
+
+    def visit_While(self, node):
+        if condition_is_unsatisfiable(node.test):
+            self._define(self.unreachable_code, 'while', node.lineno,
+                size=lines.count_lines(node),
+                message="unsatisfiable 'while' condition",
+                confidence=DEFAULT_CONFIDENCE)
+
 
     def visit(self, node):
         method = 'visit_' + node.__class__.__name__
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -96,20 +96,6 @@ def _ignore_variable(filename, varname):
         _is_special_name(varname))
 
 
-def condition_is_unsatisfiable(condition):
-    """
-    Checks if the given condition is unsatisfiable 
-    """
-    try:
-        cond = condition if isinstance(condition, ast.Expression) else ast.Expression(condition)
-        test = compile(cond, filename="<string>", mode='eval')
-        satisfiable = bool(eval(test, {}, {}))
-    except (NameError, ValueError):
-        return False
-    else:
-        return not satisfiable
-
-
 class Item(object):
     """
     Hold the name, type and location of defined code.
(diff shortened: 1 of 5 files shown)


vulture/core.py
```diff
@@ -96,6 +96,20 @@ def _ignore_variable(filename, varname):
...
 
+def condition_is_unsatisfiable(condition):
+    """
+    Checks if the given condition is unsatisfiable 
+    """
+    try:
+        cond = condition if isinstance(condition, ast.Expression) else ast.Expression(condition)
+        test = compile(cond, filename="<string>", mode='eval')
+        satisfiable = bool(eval(test, {}, {}))
+    except (NameError, ValueError):
```
jendrikseipp: We also need to catch TypeError. We need test for all errors, please add tests with comments about which errors they trigger:

    'chr(-1)'   # raises ValueError
    'getattr(True, "foo")'  # raises AttributeError
    'isinstance(True, True)  # raises TypeError

'foo' and 'foo and False' raise NameError.

vulture/core.py
```diff
@@ -425,6 +448,15 @@ def is_identifier(s):
                 if is_identifier(attr):
                     self.used_attrs.add(attr)
 
+
+    def visit_While(self, node):
+        if condition_is_unsatisfiable(node.test):
+            self._define(self.unreachable_code, 'while', node.lineno,
+                size=lines.count_lines(node),
+                message="unsatisfiable 'while' condition",
+                confidence=DEFAULT_CONFIDENCE)
```
RJ722: It is unable to calculate conditions based on dynamic variables, for ex.: `if __name__ == '__main__'`, I am not sure, but there would be a similar case with env variables.

Can we use 90% instead?

vulture/core.py
```diff
@@ -96,6 +96,20 @@ def _ignore_variable(filename, varname):
         _is_special_name(varname))
 
 
+def condition_is_unsatisfiable(condition):
+    """
+    Checks if the given condition is unsatisfiable
```
RJ722: Can we move the function to utils?

vulture/core.py
RJ722: This also uncovers an interesting thing, how are we gonna whitelist false positives. I think, it would need a no check comment (like `#noqa`).

jendrikseipp: There mustn't be any false positives for unsatisfiable conditions. In other words, we should only report cases with 100% confidence. 

The `if __name__ == '__main__'` is a good catch! We need to overwrite the `"__builtins__"` key in the globals() dict passed to eval():

Outside of try-block:
```
safe_builtins = dict((key, value) for key, value in vars(__builtins__).items() 
    if not (key.startswith('__') and key.endswith('__')) and 
    key not in set(['input', 'raw_input']))}
```
and then
```
satisfiable = bool(eval(test, {'__builtins__': safe_builtins}, {}))
```

The combined list of built-in names of Python 2 and 3 are:

    ['ArithmeticError', 'AssertionError', 'AttributeError', 'BaseException', 'BlockingIOError', 'BrokenPipeError', 'BufferError', 'BytesWarning', 'ChildProcessError', 'ConnectionAbortedError', 'ConnectionError', 'ConnectionRefusedError', 'ConnectionResetError', 'DeprecationWarning', 'EOFError', 'Ellipsis', 'EnvironmentError', 'Exception', 'False', 'FileExistsError', 'FileNotFoundError', 'FloatingPointError', 'FutureWarning', 'GeneratorExit', 'IOError', 'ImportError', 'ImportWarning', 'IndentationError', 'IndexError', 'InterruptedError', 'IsADirectoryError', 'KeyError', 'KeyboardInterrupt', 'LookupError', 'MemoryError', 'NameError', 'None', 'NotADirectoryError', 'NotImplemented', 'NotImplementedError', 'OSError', 'OverflowError', 'PendingDeprecationWarning', 'PermissionError', 'ProcessLookupError', 'RecursionError', 'ReferenceError', 'ResourceWarning', 'RuntimeError', 'RuntimeWarning', 'StandardError', 'StopAsyncIteration', 'StopIteration', 'SyntaxError', 'SyntaxWarning', 'SystemError', 'SystemExit', 'TabError', 'TimeoutError', 'True', 'TypeError', 'UnboundLocalError', 'UnicodeDecodeError', 'UnicodeEncodeError', 'UnicodeError', 'UnicodeTranslateError', 'UnicodeWarning', 'UserWarning', 'ValueError', 'Warning', 'ZeroDivisionError', '__build_class__', '__debug__', '__doc__', '__import__', '__loader__', '__name__', '__package__', '__spec__', 'abs', 'all', 'any', 'apply', 'ascii', 'basestring', 'bin', 'bool', 'buffer', 'bytearray', 'bytes', 'callable', 'chr', 'classmethod', 'cmp', 'coerce', 'compile', 'complex', 'copyright', 'credits', 'delattr', 'dict', 'dir', 'divmod', 'enumerate', 'eval', 'exec', 'execfile', 'exit', 'file', 'filter', 'float', 'format', 'frozenset', 'getattr', 'globals', 'hasattr', 'hash', 'help', 'hex', 'id', 'input', 'int', 'intern', 'isinstance', 'issubclass', 'iter', 'len', 'license', 'list', 'locals', 'long', 'map', 'max', 'memoryview', 'min', 'next', 'object', 'oct', 'open', 'ord', 'pow', 'print', 'property', 'quit', 'range', 'raw_input', 'reduce', 'reload', 'repr', 'reversed', 'round', 'set', 'setattr', 'slice', 'sorted', 'staticmethod', 'str', 'sum', 'super', 'tuple', 'type', 'unichr', 'unicode', 'vars', 'xrange', 'zip']

I don't see any other "dynamic" item, i.e., one that might change its meaning or value.

vulture/core.py
jendrikseipp: OK, good idea!

vulture/core.py
RJ722: Thank you so much!
I had not seen the usage of `__builtins__` before.

vulture/utils.py
```diff
@@ -10,6 +11,27 @@ class VultureInputException(Exception):
     pass
 
 
+def condition_is_unsatisfiable(condition):
+    """
+    Try to evaluate the given condition. Return true, if the evaluation
+    doesn't need the values of any variables and evaluates to False.
+    """
+    builtins = (vars(__builtins__) if hasattr(__builtins__, '__dict__') else
+                __builtins__)
```
jendrikseipp: __builtins__ is actually an implementation detail (note the s at the end). Let's use the following at the top of the file (https://docs.python.org/2/library/__builtin__.html, https://docs.python.org/3/library/builtins.html):
```
try:
    # Python 3
    import builtins
except ImportError:
    # Python 2
    import __builtin__ as builtins
```

Then below we need to use `vars(builtins).items()`.

jendrikseipp:
Nice! This is a cool feature to have!

Issue #90, opened 2017-08-12
unreachable else


RJ722:
Detect unreachable else clauses by detecting if True.

For example:
```python
if True:
    foo()
else:
    bar()
```

jendrikseipp:
Fixed by #91.

PR #91, opened 2017-08-12
Catch unsatisfiable 'if' conditions

RJ722:
## Related Issue
Partially handles #88 

## Checklist:


vulture/core.py
```diff
@@ -384,6 +384,14 @@ def visit_FunctionDef(self, node):
...
                 self._define_variable(param, node.lineno, confidence=100)
 
+    def visit_If(self, node):
+        if utils.condition_is_unsatisfiable(node.test):
+            self._define(self.unreachable_code, 'if', node.lineno,
+                         size=lines.count_lines(node) if self.sort_by_size
+                         else 1,
+                         message="unsatisfiable 'if' condition",
+                         confidence=100)
+
```
RJ722: TODO: We need to generalise this. How about adding `self.define_unreachable` in a different PR?

jendrikseipp: Let's first get this PR done and see if we have to reduce code duplication later.

jendrikseipp: What's important here are the tests: 

```
if False:
    unreachable
else:
     reachable
```

```
if foo:
    reachable
elif False:
    unreachable
else:
     reachable
```
-> If a test is unsatisfiable, we report it.

```
if True:
    reachable
else:
     unreachable
```

```
if True:
    reachable
elif foo:
    unreachable
else:
     unreachable
```

```
if foo:
    reachable
elif True:
    reachable
else:
     unreachable
```

-> If a test is always satisfied, all cases after it are dead code.

jendrikseipp: You will probably need to convert is_condition_unsatisfiable() to evaluate_condition() and let it return True, False or raise ValueError if the condition cannot be evaluated by ast.literal_eval().

tests/test_condition_eval.py
```diff
@@ -0,0 +1,129 @@
...
+    check_unreachable(v, 2, 2, 'while')
+
+
+def test_while_nested(v):
+    v.scan("""\
+def foo():
+    while True:
+        while False:
+            pass
+""")
```
jendrikseipp: check is missing

tests/test_condition_eval.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def test_if_false(v):
+    v.scan("""\
+def foo():
```
jendrikseipp: let's remove `def foo()` where it's not needed to make the code clearer.

tests/test_condition_eval.py
```diff
@@ -0,0 +1,129 @@
...
+        if bar():
+            pass
+        elif False:
+            print("Unreachable")
+        else:
+            pass
+    else:
+        pass
+""")
+    check_unreachable(v, 5, 4, 'if')
```
jendrikseipp: the unreachable code should have size 2 only.

tests/test_condition_eval.py
```diff
@@ -0,0 +1,129 @@
...
+        if bar():
+            pass
+        elif True:
+            print("Unreachable")
+        else:
+            pass
+    else:
+        pass
+""")
+    check_unreachable(v, 5, 4, 'if')
```
jendrikseipp: The dead code is in lines 7-8, not 5-8.

vulture/utils.py
```diff
@@ -11,7 +11,7 @@ class VultureInputException(Exception):
     pass
 
 
-def condition_is_unsatisfiable(condition):
+def evaluate_condition(condition):
     """
     Try to safely evaluate the given condition. Return true, if the
     evaluation succeeds and the condition evaluates to False.
```
jendrikseipp: Update docstring.

tests/test_condition_eval.py
```diff
@@ -0,0 +1,129 @@
...
+        check_condition("False and 1", False)
+        check_condition("False and foo", False)
+        check_condition('chr(-1)', False)
+        # checks AttributeError is handled properly
+        check_condition('getattr(True, "foo")', False)
+        check_condition("hasattr(str, 'foo')", False)
+        # checks TypeError is handled properly
+        check_condition('isinstance(True, True)', False)
+
+        check_condition("not False", False)
```
jendrikseipp: I think this will only test the first item ("foo"). We probably need a for loop which loops over the strings ["foo", "foo and False", ...].

tests/test_condition_eval.py
```diff
@@ -0,0 +1,129 @@
+import ast
```
jendrikseipp: please rename the file to test_conditions.py

tests/test_condition_eval.py
RJ722: The main advantage of using `eval` over `ast.literal_eval` is we can catch statements with multiple conditions - That's what I think is the main cause when a user might generally neglect an `if` condition.

jendrikseipp: I'll look into that. However, the solution should be based on ast.literal_eval(), not eval().

tests/test_condition_eval.py
RJ722: Sorry! :p

vulture/utils.py
```diff
@@ -11,10 +11,11 @@ class VultureInputException(Exception):
     pass
 
 
-def condition_is_unsatisfiable(condition):
+def evaluate_condition(condition):
```
jendrikseipp: It's probably best to add two new functions that call this function: `condition_is_always_true` and `condition_is_always_false`.

vulture/core.py
```diff
@@ -384,6 +384,33 @@ def visit_FunctionDef(self, node):
             if param and isinstance(param, str):
                 self._define_variable(param, node.lineno, confidence=100)
 
+    def visit_If(self, node):
+        try:
+            if utils.evaluate_condition(node.test):
+                dead_node = getattr(node, 'orelse')
```
jendrikseipp: dead_node -> else_body

jendrikseipp: define else_body above `if` to avoid defining it twice.

tests/test_conditions.py
```diff
@@ -0,0 +1,126 @@
...
+                  'False and foo',
+                  'chr(-1)',
+                  'getattr(True, "foo")',
+                  'hasattr(str, "foo")',
+                  'isinstance(True, True)',
+                  'not False']
+    for condition in conditions:
+        with pytest.raises(ValueError) as e:
+            check_condition(condition, False)
+        assert e.match("Condition cannot be evaluated")
```
jendrikseipp: should this be indented?

vulture/core.py
```diff
@@ -384,6 +384,32 @@ def visit_FunctionDef(self, node):
             if param and isinstance(param, str):
                 self._define_variable(param, node.lineno, confidence=100)
 
+    def visit_If(self, node):
+        try:
+            if utils.evaluate_condition(node.test):
+                orelse = getattr(node, 'orelse')
```
jendrikseipp: after all `orelse` is fine as variable name :-)

RJ722: Sorry, I didn't see your comment until now. Changed it to `else_body` already.

tests/test_conditions.py
RJ722: I preferred it because [The official pytest documentation](https://docs.pytest.org/en/latest/assert.html) uses it.
Should we?

RJ722:
Closes #90

jendrikseipp:
Thanks!

RJ722:
It is such a pleasure working with you! Thanks! 🎉

PR #94, opened 2017-08-28
Add Item.get_report method.

### LINKED ISSUE #92: Filtering items with a regex
I currently have [a custom script](https://github.com/qutebrowser/qutebrowser/blob/master/scripts/dev/run_vulture.py) to run vulture for two reasons:

- Dynamically generating a whitelist.
- Filtering any functions matching `[a-z]+[A-Z][a-zA-Z]+` because those are typically virtual methods provided by Qt/PyQt, i.e. from C++ - any Python methods are named `snake_case` in my project.

Unfortunately, the script broke quite often recently because things changed in vulture in subtle ways. I don't think the whitelist is the problem (after all, I just pass that to vulture), but the filtering means I need to have my own `report` there so I can filter the items before they are reported.

I say two ways around this:

- Adding something like a `--func-filter-regex` (and maybe `--props-filter-regex`, etc.) commandline argument.
- Adding something like a `filter_callback` to `vulture.report`, which is only intended for when using vulture programmatically, and gets called with each `Item` once. If it returns `True`, the given item does not get reported.

Both of those would probably make my script a bit less relying on vulture internals. For the first one, I could even switch to running `vulture.main`.

What do you think?

RJ722:
Closes #92 

## Checklist:

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -116,6 +116,16 @@ class Item(object):
         assert self.last_lineno >= self.first_lineno
         return self.last_lineno - self.first_lineno + 1
 
+    def get_report(self, add_size=False):
+        if add_size:
+            line_format = 'line' if self.size == 1 else 'lines'
+            size_report = ', {0:d} {1}'.format(self.size, line_format)
+        else:
+            size_report = ''
+        return "{0}:{1:d}: {2} ({3}% confidence{4})".format(
+                utils.format_path(self.filename), self.first_lineno,
+                self.message, self.confidence, size_report)
+
     def _tuple(self):
         return (self.filename, self.first_lineno, self.name)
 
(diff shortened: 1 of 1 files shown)


RJ722:
CC @The-Compiler

vulture/core.py
```diff
@@ -116,6 +116,17 @@ def size(self):
         assert self.last_lineno >= self.first_lineno
         return self.last_lineno - self.first_lineno + 1
 
+    def get_report(self, sort_by_size=False):
```
jendrikseipp: sort_by_size -> add_size (there's no sorting inside get_report)

vulture/core.py
```diff
@@ -116,6 +116,17 @@ def size(self):
...
+        if sort_by_size:
+            line_format = 'line' if self.size == 1 else 'lines'
+            size_report = ', {0:d} {1}'.format(self.size, line_format)
+        else:
+            size_report = ''
+        message = "{0}:{1:d}: {2} ({3}% confidence{4})".format(
```
jendrikseipp: no need for `message` variable, just return it.

RJ722: Done!

jendrikseipp:
Thanks! I'll make a new release shortly.

PR #97, opened 2017-09-25
Add AppVeyor badge to README

hugovk:
## Description

Show the build status of both CIs clearly on the README/

## Checklist:

- [n/a] I have added tests to cover my changes.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,7 +5,7 @@ Vulture - Find dead code
    :target: https://travis-ci.org/jendrikseipp/vulture
    :alt: Travis CI build status (Linux)
 
-.. image:: https://img.shields.io/appveyor/ci/jendrikseipp96693/vulture/master.svg?label=Windows%20build
+.. image:: https://ci.appveyor.com/api/projects/status/github/jendrikseipp/vulture?svg=true
    :target: https://ci.appveyor.com/project/jendrikseipp96693/vulture
    :alt: AppVeyor CI build status (Windows)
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1,8 +1,13 @@
 Vulture - Find dead code
 ========================
 
-.. image:: https://travis-ci.org/jendrikseipp/vulture.svg?branch=master
+.. image:: https://travis-ci.org/jendrikseipp/vulture.svg?branch=master&label=Linux%20build
```
jendrikseipp: It seems the label parameter is ignored. Is that so? If yes, I think it should be removed.

hugovk: Oops, I meant to also serve this from https://img.shields.io/travis/jendrikseipp/vulture.svg?branch=master&label=Linux%20build so the parameter would be used. 

Which would you prefer: remove the parameter or serve from same place?

jendrikseipp: I think it would be best if we could avoid using another service (i.e., shields.io). Can we use https://www.appveyor.com/docs/status-badges/ directly? The titles are not so important.

hugovk: I've removed the Travis parameter.

I can't see the AppVeyor badge URL, they hide it away:

> You can see a badge image URL and sample markup on the Badges tab of project settings. Badge URLs for both private and public projects contain a security token.

Please can you paste the URL here or update the PR directly?

jendrikseipp:
I found a link that doesn't need the security token. Thanks!

PR #99, opened 2017-10-08
DOC: update README; `vulture` has multiple modules

johnyf:
`vulture` used to be a single module, but is now a package with multiple modules.


jendrikseipp:
Thanks!

Issue #100, opened 2017-10-26
Detect while True: else statements


RJ722:
vulture currently detects the case when a `else` block is associated with an `if True` statement, but does not do so in the case of a `while` loop.

For example, the following example wouldn't be detected as unused code:
```py
while True:
    print("Hi")
else:
    # Dead code
    print("Dead code")
```

jendrikseipp:
Good catch!

RJ722:
I would like to submit a PR. 😄

jendrikseipp:
Perfect!

PR #102, opened 2017-10-29
Report while-True-else block as unreachable

RJ722:
When a while loop encounters a `True` statement, report `else` block as unreachable.

Closes: #100 

## Checklist:


jendrikseipp:
Perfect!

PR #114, opened 2018-05-31
README.rst: Add examples and how-to's for setup

RJ722:
Add examples output for Vulture.

## Checklist:

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,9 @@ commands =
   py.test
 
 [testenv:docs]
-deps = collective.checkdocs
+deps =
+    collective.checkdocs
+    pygments
 commands = python setup.py checkdocs
 
 [testenv:style]
--- a/README.rst
+++ b/README.rst
@@ -101,90 +101,58 @@ dead code first.
 Examples
 --------
 
-**Analysis**
-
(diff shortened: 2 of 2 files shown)


RJ722:
Would it be a nice idea to also mention little snippets of `.travis.yml` or `appveyor.yml` showing how to add Vulture to CI tests?

tox.ini
```diff
@@ -16,7 +16,9 @@ commands =
   py.test
 
 [testenv:docs]
-deps = collective.checkdocs
+deps =
+    collective.checkdocs
+    pygments
```
jendrikseipp: Why do we need pygments?

jendrikseipp:
Thanks for looking into this! I think instead of snippets for CI, we should document the exit codes. Let me make some changes to the patch.

tox.ini
RJ722: The `pygments` library is required for syntax highlighting.

jendrikseipp: Ah, I see, thanks!

RJ722:
> we should document the exit codes

Yes, that is a great idea. I'm gonna make a PR.

PR #116, opened 2018-06-01
Ignore pytest_cache and swp files

RJ722:
## Description

Git keeps on catching `pytest_cache/` (because of `tox`) and `.swp` files (because of `vim`). It would be great if we could ignore these.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ vulture.egg-info/
 .cache/
 .coverage
 .tox/
+.pytest_cache/
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,5 @@ htmlcov/
 vulture.egg-info/
 .cache/
 .coverage
-.tox/
 .pytest_cache/
+.tox/


.gitignore
```diff
@@ -7,3 +7,5 @@ vulture.egg-info/
 .cache/
 .coverage
 .tox/
+.pytest_cache/
```
jendrikseipp: Which command in our tests produces this directory? I can't generate this with "tox".

.gitignore
```diff
@@ -7,3 +7,5 @@ vulture.egg-info/
...
+*.swp
```
jendrikseipp: Can you add this ignore in your global .gitignore file?

.gitignore
RJ722: I'm not sure about what command generates it, but yeah every time I run `tox`, it is there. 😛

.gitignore
RJ722: Sure!

.gitignore
jendrikseipp: OK, my tox version was older. Now I see it. This should indeed be added to .gitignore. Can you remove the *.swp ignore, please?

RJ722: Done!

jendrikseipp:
Thanks!

PR #119, opened 2018-06-02
Use argparse instead of optparse

RJ722:
## Checklist:

### DIFF
--- a/tests/test_script.py
+++ b/tests/test_script.py
@@ -50,3 +50,14 @@ def test_min_confidence():
     assert call_vulture([
         'vulture/core.py', '--exclude', 'whitelists',
         '--min-confidence', '100']) == 0
+
+
+def test_exclude():
+    def mix_path(filenames):
+        return ",".join(
+            [os.path.join('vulture', filename) for filename in filenames])
+    assert call_vulture([
+        'vulture/', '--exclude', mix_path(['core.py', 'utils.py'])]) == 1
+    assert call_vulture([
+        'vulture/', '--exclude',
+        mix_path(['core.py', 'utils.py', 'lines.py'])]) == 0
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -26,9 +26,9 @@
 
 from __future__ import print_function
 
+import argparse
 import ast
 from fnmatch import fnmatchcase
-import optparse
 import os.path
 import pkgutil
 import re
@@ -491,40 +491,36 @@ class Vulture(ast.NodeVisitor):
 
 
 def _parse_args():
-    def csv(option, _, value, parser):
-        setattr(parser.values, option.dest, value.split(','))
-    usage = """\
-usage: %prog [options] PATH [PATH ...]
-
-Paths may be Python files or directories. For each directory vulture
-analyzes all contained *.py files.
-"""
+    def csv(exclude):
+        return exclude.split(',')
+    usage = "%(prog)s [options] PATH [PATH ...]"
     version = "vulture {0}".format(__version__)
-    parser = optparse.OptionParser(usage=usage, version=version)
-    parser.add_option(
-        '--exclude', action='callback', callback=csv, metavar='PATTERN',
-        type='string', default=[],
-        help=(
-            'Comma-separated list of paths to ignore (e.g.,'
-            ' *settings.py,docs/*.py). PATTERNs can contain globbing'
-            ' characters (*, ?, [, ]). Treat PATTERNs without globbing'
-            ' characters as *PATTERN*.'))
-    parser.add_option(
-        "--sort-by-size", action="store_true",
-        help="Sort unused functions and classes by their lines of code")
-    parser.add_option('-v', '--verbose', action='store_true')
-    parser.add_option(
-        '--min-confidence', action='store', type='int', default=0, help=(
-            'Minimum confidence (between 0 and 100) for code to be'
-            ' reported as unused.'))
-    options, args = parser.parse_args()
-    return options, args
+    parser = argparse.ArgumentParser(prog='vulture', usage=usage)
+    parser.add_argument(
+        'paths', action='store', nargs='+', metavar='PATH', help='Paths'
+        ' may be Python files or directories. For each directory'
+        ' Vulture analyzes all contained *.py files.')
+    parser.add_argument(
+        '--exclude', metavar='PATTERN', type=csv, help='Comma-separated list'
+        ' of paths to ignore (e.g., *settings.py,docs/*.py). PATTERNs can'
+        ' contain globbing characters (*, ?, [, ]). Treat PATTERNs without'
+        ' globbing characters as *PATTERN*.')
+    parser.add_argument(
+        '--min-confidence', action='store', type=int, default=0, help='Minimum'
+        ' confidence (between 0 and 100) for code to be'
+        ' reported as unused.')
+    parser.add_argument(
+        "--sort-by-size", action="store_true", help='Sort'
+        ' unused functions and classes by their lines of code')
+    parser.add_argument('-v', '--verbose', action='store_true')
+    parser.add_argument('--version', action='version', version=version)
+    return parser.parse_args()
 
 
 def main():
-    options, args = _parse_args()
-    vulture = Vulture(verbose=options.verbose)
-    vulture.scavenge(args, exclude=options.exclude)
+    args = _parse_args()
+    vulture = Vulture(verbose=args.verbose)
+    vulture.scavenge(args.paths, exclude=args.exclude)
     sys.exit(vulture.report(
-        min_confidence=options.min_confidence,
-        sort_by_size=options.sort_by_size))
+        min_confidence=args.min_confidence,
+        sort_by_size=args.sort_by_size))
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -4,6 +4,7 @@ News
 0.27 (2018-06-05)
 -----------------
 * Report ``while (True): ... else: ...`` as unreachable (thanks @RJ722).
+* Use ``argparse`` instead of ``optparse``.
 * Whitelist Mock.return_value and Mock.side_effect in unittest.mock module.
 * Drop support for Python 2.6 and 3.3.
 * Improve documentation and test coverage (thanks @RJ722).
--- a/TODO.rst
+++ b/TODO.rst
@@ -4,7 +4,6 @@ TODOs
 * Uuse coverage.py to detect false-positives (#109).
 * Write script for turning SIP files into whitelists (#8).
 * Add PyQT whitelist to Vulture repo.
-* Use argparse instead of optparse.
 
 
 Non-TODOs


vulture/core.py
```diff
@@ -489,40 +489,35 @@ def generic_visit(self, node):
...
 def _parse_args():
-    def csv(option, _, value, parser):
-        setattr(parser.values, option.dest, value.split(','))
-    usage = """\
-usage: %prog [options] PATH [PATH ...]
-
-Paths may be Python files or directories. For each directory vulture
-analyzes all contained *.py files.
-"""
+    usage = "%(prog)s PATH [PATH ...] [options]"
```
jendrikseipp: Please leave [options] before PATH.

vulture/core.py
```diff
@@ -489,40 +489,35 @@ def generic_visit(self, node):
...
-    parser.add_option(
-        '--min-confidence', action='store', type='int', default=0, help=(
+    parser = argparse.ArgumentParser(prog='vulture', usage=usage)
+    parser.add_argument(
+        'paths', action='store', nargs='*', metavar='PATH', help=(
+            'Paths may be Python files or directories. For each directory'
+            ' vulture analyzes all contained *.py files.'))
+    parser.add_argument(
+        '--exclude', action='store', metavar='PATTERN', type=str, default=[],
+        nargs="+", help=(
```
jendrikseipp: `nargs='*'`, I guess?

vulture/core.py
```diff
@@ -489,40 +489,35 @@ def generic_visit(self, node):
...
+            'Paths to ignore (e.g., *settings.py docs/*.py). PATTERNs can'
```
jendrikseipp: I think that `vulture --exclude *settings.py docs/*.py myprogram.py`  would check (and not ignore) the docs folder.

vulture/core.py
```diff
@@ -489,40 +489,35 @@ def generic_visit(self, node):
...
-            ' characters as *PATTERN*.'))
-    parser.add_option(
-        "--sort-by-size", action="store_true",
-        help="Sort unused functions and classes by their lines of code")
-    parser.add_option('-v', '--verbose', action='store_true')
-    parser.add_option(
-        '--min-confidence', action='store', type='int', default=0, help=(
+    parser = argparse.ArgumentParser(prog='vulture', usage=usage)
+    parser.add_argument(
+        'paths', action='store', nargs='*', metavar='PATH', help=(
```
jendrikseipp: `nargs='+'`

vulture/core.py
```diff
@@ -489,40 +489,35 @@ def generic_visit(self, node):
...
             'Minimum confidence (between 0 and 100) for code to be'
             ' reported as unused.'))
-    options, args = parser.parse_args()
-    return options, args
+    parser.add_argument("--sort-by-size", action="store_true", help=(
+        'Sort unused functions and classes by their lines of code'))
+    parser.add_argument('-v', '--verbose', action='store_true')
+    parser.add_argument('--version', action='version', version=version)
+    args = parser.parse_args()
+    return args
```
jendrikseipp: Use `return parser.parse_args()`

vulture/core.py
jendrikseipp: I think we should add tests for this by listing excludes not as `--exclude tests/*`, but `--exclude tests/myfile.py tests/mysecondfile.py` in one test. (Choose files appropriately.)

vulture/core.py
RJ722: If we use `*`, then `vulture core/ --exclude` will be a valid command. Do you think that we should throw an error in such a situation?

vulture/core.py
RJ722: The command `vulture --exclude *settings.py docs/*.py myprogram.py` would just exclude all the three because we are taking up any number of arguments occurring after `--exclude`. This is also the reason why I've placed PATH before [options].

vulture/core.py
RJ722: Discussed in https://github.com/jendrikseipp/vulture/pull/119#discussion_r192597306

vulture/core.py
RJ722: > I think we should add tests for this by listing excludes not as --exclude tests/*, but --exclude tests/myfile.py tests/mysecondfile.py in one test. (Choose files appropriately.)

Yes, I see that we aren't explicitly testing `--exclude`. I am gonna add a test.


One more thing I've been wondering: If we use glob patterns, wouldn't shell automatically expand it to the list of all matching files even before passing it on to vulture. This can have unwanted side effects - Let's take up the example `vulture vulture/ --exclude *settings.py` - If no file matches the glob `*settings.py`, shell would automatically abort the process before even starting to execute Vulture. I think the expected behavior should be to silently ignore the glob if no files are found.

vulture/core.py
RJ722: Yes, that should be the expected behavior, but I saw that originally vulture did not throw an error if run without any arguments - I just thought that it would be nice to continue it. Anyways, changing it now.

RJ722: Also, do we need to document it somewhere?

RJ722:
I can't think of an obvious reason for appveyor failing.

The command `vulture vulure/ --exclude vulture/core.py` is supposed to give the following output:
```
vulture/lines.py:40: unused function 'get_last_line_number' (60% confidence)
vulture/utils.py:44: unused function 'condition_is_always_false' (60% confidence)
vulture/utils.py:48: unused function 'condition_is_always_true' (60% confidence)
vulture/utils.py:52: unused function 'format_path' (60% confidence)
vulture/utils.py:78: unused function 'read_file' (60% confidence)
vulture/utils.py:98: unused class 'LoggingList' (60% confidence)
vulture/utils.py:110: unused class 'LoggingSet' (60% confidence)
```

vulture/core.py
jendrikseipp: No, I think we should abort in that case. Vulture just cannot distinguish between a list generated by the user or a shell-glob.

vulture/core.py
jendrikseipp: Ah, ok. Then we should indeed use `nargs='+'`. But this doesn't solve the problem of distinguishing between excluded and included paths. I think the best solution is to accept a single comma-separated list of strings for the `--exclude` argument as before. This removes the ambiguity and is backwards-compatible.

tests/test_script.py
```diff
@@ -53,3 +53,7 @@ def test_min_confidence():
     assert call_vulture([
         'vulture/core.py', '--exclude', 'whitelists',
         '--min-confidence', '100']) == 0
+
+
+def test_exclude():
+    assert call_vulture(['vulture/', '--exclude', 'vulture/core.py']) == 1
```
jendrikseipp: What I meant was to exclude multiple individual files, i.e., in a comma-separated list.

vulture/core.py
jendrikseipp: This is documented by "PATH [PATH ...]" and I don't think we need to put it into the README.

RJ722: Ah, correct!

vulture/core.py
```diff
@@ -491,40 +491,35 @@ def generic_visit(self, node):
...
+    return parser.parse_args()
 
 
 def main():
-    options, args = _parse_args()
-    vulture = Vulture(verbose=options.verbose)
-    vulture.scavenge(args, exclude=options.exclude)
+    args = _parse_args()
+    vulture = Vulture(verbose=args.verbose)
+    exclude = [str(excluded) for excluded in args.exclude.split(',')]
```
jendrikseipp: Let's use the `type` parameter to pass a conversion function: https://docs.python.org/3/library/argparse.html#type

tests/test_script.py
```diff
@@ -50,3 +59,12 @@ def test_min_confidence():
     assert call_vulture([
         'vulture/core.py', '--exclude', 'whitelists',
         '--min-confidence', '100']) == 0
+
+
+def test_exclude():
+    expected_output = """\
+vulture/lines.py:40: unused function 'get_last_line_number' (60% confidence)
```
jendrikseipp: Checking the output is brittle. Whenever we add or remove a line before line 40, we need to adapt this test.

Can we just check the exit code instead?

`['vulture/', '--exclude', 'vulture/core.py,vulture/utils.py']` -> 1
`['vulture/', '--exclude', 'vulture/core.py,vulture/utils.py,vulture/lines.py']` -> 0

vulture/core.py
```diff
@@ -491,40 +491,36 @@ def generic_visit(self, node):
...
-    def csv(option, _, value, parser):
-        setattr(parser.values, option.dest, value.split(','))
-    usage = """\
-usage: %prog [options] PATH [PATH ...]
-
-Paths may be Python files or directories. For each directory vulture
-analyzes all contained *.py files.
-"""
+    def csv(exclude):
+        return [str(excluded) for excluded in exclude.split(',')]
```
jendrikseipp: I think we can just `return exclude.split()`.

vulture/core.py
```diff
@@ -491,40 +491,36 @@ def generic_visit(self, node):
...
-    parser = optparse.OptionParser(usage=usage, version=version)
-    parser.add_option(
-        '--exclude', action='callback', callback=csv, metavar='PATTERN',
-        type='string', default=[],
-        help=(
+    parser = argparse.ArgumentParser(prog='vulture', usage=usage)
+    parser.add_argument(
+        'paths', action='store', nargs='+', metavar='PATH', help=(
+            'Paths may be Python files or directories. For each directory'
+            ' vulture analyzes all contained *.py files.'))
```
jendrikseipp: vulture -> Vulture

vulture/core.py
```diff
@@ -491,40 +491,36 @@ def generic_visit(self, node):
...
-    parser.add_option('-v', '--verbose', action='store_true')
-    parser.add_option(
-        '--min-confidence', action='store', type='int', default=0, help=(
+    parser.add_argument(
+        '--min-confidence', action='store', type=int, default=0, help=(
             'Minimum confidence (between 0 and 100) for code to be'
             ' reported as unused.'))
-    options, args = parser.parse_args()
-    return options, args
+    parser.add_argument("--sort-by-size", action="store_true", help=(
```
jendrikseipp: Are the brackets for the four `help=()` calls really needed?

RJ722: Given the other option is to use a `\`, I'd say that we should continue using brackets.

Source: https://stackoverflow.com/a/5437652/6426752

vulture/core.py
RJ722: Yes and yes. 😲

vulture/core.py
jendrikseipp: The brackets are not needed since the strings occur inside brackets anyway.

RJ722: Yes, you are right! (as always) :-)

jendrikseipp:
Thanks!

jendrikseipp:
The fnmatch module docs state "Note that the filename separator ('/' on Unix) is not special to this module.". Therefore, users can just use / or \ as they please.

RJ722:
How did the fix for windows even work?? - I thought that `vulture\core.py` wasn't matching against `vulture/core.py` 😛

PR #121, opened 2018-06-02
Test Vulture for non utf-8 encoding

RJ722:
## Checklist:

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,3 +1,6 @@
+from codecs import BOM_UTF16_LE
+import os
+
 import pytest
 
 from . import v
@@ -21,3 +24,21 @@ def foo():
 """)
     with pytest.raises(ValueError):
         v.get_unused_code(min_confidence=150)
+
+
+def test_non_utf8_encoding(v):
+    def _write_bom_encoded_utf_16_le(f, content):
+        f.write(BOM_UTF16_LE)
+        f.write(content.encode('utf_16_le'))
+        f.flush()  # Ensure that the file is actually written to disk.
+    code = """\
+def foo():
+    pass
+
+foo()
+"""
+    with open('non-utf8.py', 'wb') as f:
+        _write_bom_encoded_utf_16_le(f, code)
+        v.scavenge([f.name])
+    os.remove(f.name)
+    assert v.found_dead_code_or_error
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,4 +1,4 @@
-from codecs import BOM_UTF16_LE
+import codecs
 import os
 
 import pytest
@@ -27,10 +27,6 @@ def foo():
 
 
 def test_non_utf8_encoding(v):
-    def _write_bom_encoded_utf_16_le(f, content):
-        f.write(BOM_UTF16_LE)
-        f.write(content.encode('utf_16_le'))
-        f.flush()  # Ensure that the file is actually written to disk.
     code = """\
 def foo():
     pass
@@ -38,7 +34,8 @@ def foo():
 foo()
 """
     with open('non-utf8.py', 'wb') as f:
-        _write_bom_encoded_utf_16_le(f, code)
-        v.scavenge([f.name])
+        f.write(codecs.BOM_UTF16_LE)
+        f.write(code.encode('utf_16_le'))
+    v.scavenge([f.name])
     os.remove(f.name)
     assert v.found_dead_code_or_error
--- a/tests/test_script.py
+++ b/tests/test_script.py
@@ -1,11 +1,7 @@
-import glob
-import os.path
 import subprocess
 import sys
 
-DIR = os.path.dirname(os.path.abspath(__file__))
-REPO = os.path.dirname(DIR)
-WHITELISTS = glob.glob(os.path.join(REPO, 'vulture', 'whitelists', '*.py'))
+from . import REPO, WHITELISTS
 
 
 def call_vulture(args, **kwargs):
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,7 +1,13 @@
 import ast
+import glob
+import os.path
 
 import pytest
 from vulture import core
+
+DIR = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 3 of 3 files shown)


jendrikseipp:
Good idea to test for this. Instead of checking in a new file, I'd create a temporary non-UTF-8 file with the `tempfile` module.

tests/test_script.py
```diff
@@ -53,3 +54,16 @@ def test_min_confidence():
...
+    with tempfile.NamedTemporaryFile(mode='wb', dir=REPO, suffix='.py') as f:
+        code = """\
+def foo():
+    pass
+
+foo()
+"""
+        f.write(code.encode("utf_16_be"))
+        f.close()
+        assert call_vulture([f.name]) == 1
```
jendrikseipp: Can we dedent the last line and remove the call `f.close()` (which should happen automatically after exiting the with-block)?

jendrikseipp:
It seems that the new test covers no new lines.

tests/test_script.py
RJ722: As soon as we are out of the `with` block, `tempfile` would delete the file created. This would cause a false negative, as exit code `1` would be because of file not found error.

jendrikseipp: OK, thanks. I think it would be best to avoid such possible confusion by using the API, not the CLI to test non-UTF8 files.

RJ722: Yes, agreed!

RJ722:
We observed no new coverage because the error code `1` wasn't because of `VultureInputException`, but the following:
```
tmp4xycftl1.py: invalid source code "source code string cannot contain null bytes"
```

After further investigation, it turned out that both `tokenize` and `codecs` (even with encoding set to `"utf_8"`) could read the `"utf_16_be"` encoded file correctly, but the `read_file` function passed it on as byte data to Vulture, thus causing the failure. API would definitely be a better option.

tests/test_script.py
jendrikseipp: I agree.

RJ722:
The previous error was because apparently `codecs` and `tokenize` only fails on signed encodings, like "UTF-16 LE with BOM" or "UTF-16 BE with BOM".

tests/test_errors.py
```diff
@@ -21,3 +24,19 @@ def foo():
...
+def foo():
+    pass
+
+foo()
+"""
+    tests = os.path.join(REPO, 'tests')
+    with tempfile.NamedTemporaryFile(mode='wb', dir=tests, suffix='.py') as f:
+        f.write(codecs.BOM_UTF16_LE)
+        f.write(code.encode('utf_16_le'))
+        f.flush()  # Ensure that the file is actually written on disk.
```
jendrikseipp: on -> to

tests/test_errors.py
```diff
@@ -1,6 +1,9 @@
+import codecs
+import os
 import pytest
```
jendrikseipp: move pytest below stdlib imports and leave empty line between them.

tests/test_errors.py
```diff
@@ -21,3 +24,19 @@ def foo():
...
+
+def test_non_utf8_encoding(v):
+    code = """\
+def foo():
+    pass
+
+foo()
+"""
+    tests = os.path.join(REPO, 'tests')
+    with tempfile.NamedTemporaryFile(mode='wb', dir=tests, suffix='.py') as f:
```
jendrikseipp: Can we just remove the `dir` parameter and the `tests` variable and let Python choose a temporary dir?

tests/test_errors.py
```diff
@@ -21,3 +24,19 @@ def foo():
...
+        f.write(codecs.BOM_UTF16_LE)
+        f.write(code.encode('utf_16_le'))
+        f.flush()  # Ensure that the file is actually written on disk.
+        v.scavenge([f.name])
+        assert v.found_dead_code_or_error
```
jendrikseipp: The last line can be dedented.

tests/test_errors.py
RJ722: Would it still be a good idea to leave `DIR`, `REPO` and `WHITELISTS` in `__init__.py`?

RJ722: > Can we just remove the dir parameter and the tests variable and let Python choose a temporary dir?

I'm trying to remember why I did that on the firsthand? 😅

jendrikseipp:
Yes, I think that's a good idea.

tests/test_errors.py
jendrikseipp: You can leave DIR, REPO and WHITELISTS in __init__.py

tests/test_errors.py
```diff
@@ -21,3 +24,18 @@ def foo():
...
+
+
+def test_non_utf8_encoding(v):
+    code = """\
+def foo():
+    pass
+
+foo()
+"""
+    with tempfile.NamedTemporaryFile(mode='wb', suffix='.py') as f:
```
jendrikseipp: I guess appveyor doesn't allow writing to the global tmp dir. Let's use `dir=REPO`.

RJ722: The tests were failing with the same error message even when we had `dir=REPO`. It is a Windows problem I guess, see https://stackoverflow.com/questions/17502704/python-tempfile-how-to-open-several-times-the-temp-file#17502778

RJ722: Edit: When we had `dir=tests`.

jendrikseipp: Hmm, strange. But I'm confident you'll figure it out :-) Maybe you can try creating the file with a simple `with open(...) as f:` call and see if that works.

RJ722: Python docs say:

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).

So yes, we should just use plain old `open` instead - We aren't doing anything fancy with `tempfile` anyways. The only additional step would be to delete the file once we are done.

jendrikseipp:
Let me make some minor adjustments.

jendrikseipp:
Thanks!

PR #124, opened 2018-06-03
Add coveralls to track code coverage

RJ722:
## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,10 +13,12 @@ matrix:
   allow_failures:
   - python: 3.7-dev
 install:
-  - pip install tox
+  - pip install tox coveralls
   - python setup.py bdist_wheel
   - pip install --use-wheel ./dist/vulture-*.whl
 script:
   # Run the installed version of vulture on local code.
   - vulture vulture/ tests/
   - tox -e cleanup,py,docs,style
+after_success:
+  - coveralls


RJ722:
Hey @jendrikseipp! I am not sure, but I guess that you'd need to enable the analysis for `jendrikseipp/vulture` on https://coveralls.io

jendrikseipp:
I'll make a commit to see if the integration works.

RJ722:
https://coveralls.io/github/jendrikseipp/vulture - It worked! \o/

jendrikseipp:
Nice! If you like, you can try raising the coverage. Keep in mind though that 100% coverage is not a goal, but having good tests is :-) It would be great if we could find simple, reasonable tests. I'd start with the simplest cases and making a pull request for these, before working on harder ones.

jendrikseipp:
Also, keep in mind that we only have to hit the red lines, not the orange ones (which are already executed by a subset of the tested interpreters).

jendrikseipp:
Agreed. I also can't remember the use for `__repr__`. Feel free to remove it.

RJ722:
I remember it now - `__repr__` was created to have meaningful output while testing (`pytest` calls it implicitly)

jendrikseipp:
Ah, ok. When does pytest call it? Can we trigger this in our tests?

RJ722:
It only calls it when our code fails so that it can display the diff properly. I think that the easiest way to trigger it would be to print an `Item`.

jendrikseipp:
Printing calls `__str__`, we need to call `assert repr(item) == "'{}'".format(item.name)` or similar.

RJ722:
I don't know the internals of `repr`, but my answer was solely based on a very small experiment I performed. 😛

Consider the following file:
```py
import vulture

code = """\
import os
"""

v = vulture.Vulture()
v.scan(code)

print(v.get_unused_code()[0])
```

If I execute this file with `__repr__` function intact, I get `'os'` as output, whereas if I remove the `__repr__` function, I get `<vulture.core.Item object at 0x100e982d0>`

May be, if print doesn't find `__str__`, it falls back to `__repr__`?

jendrikseipp:
See my previous comment. I think it should suffice to put that snippet in a suitable place in the tests (the snippet is untested so you may have to fix it).

jendrikseipp:
After looking at the coveralls report again, I think 100% test coverage is achievable after all :-) I think almost all remaining lines can be covered with suitable tests. The special case for Python 3.3 in the _define method can be removed. And we have to find out why visit_AsyncFunctionDef is not tested. We also need to find out how `module_data` in the `scavenge` method can become `None`. If you can't trigger this problem, we can remove the `if module_data is None` check.

RJ722:
> I think 100% test coverage is achievable after all :-) 

Hooray! \o/

> And we have to find out why `visit_AsyncFunctionDef` is not tested.

That is rather strange! 😕 

> if module_data is None

I think it remedied `python2.6`, but I'm not sure.

RJ722:
> I removed the module_data check and the special case for Python 3.3 

Hooray! 100% code coverage! 🎉 \o/ 
> will leave the last untested part of the code (visit_AsyncFunctionDef) to you :-)

That one was weirdly interesting - I've taken a stab at it in #128! :-)

PR #126, opened 2018-06-04
Improve tests

### LINKED ISSUE #125: Test --sort-by-size
As discussed in #124, the test for `--sort-by-size` is faulty because it never actually calls the underlying function as no dead code is detected on the first hand.

```py
def test_sort_by_size():
    assert call_vulture(['vulture/', '--sort-by-size']) == 0
```

We'd be needing a completely new kind of test - where we can match the output of vulture to the expected output. This way, we can also test for `verbose` mode.

RJ722:
## Description
Test for `sort-by-size`. Add tests to ensure that proper error is raised when `min_confidence` is not in range of [0, 100].

## Related Issue
Closes #125

## Checklist:

### DIFF
--- a/tests/test_sorting.py
+++ b/tests/test_sorting.py
@@ -1,0 +1,17 @@
+from . import v
+assert v  # Silence pyflakes
+
+
+def test_sorting(v):
+    v.scan("""\
+def foo():
+    print("Hello, I am a long function.")
+    return "World"
+
+def bar():
+    pass
+""")
+    assert [item.name for item in
+            v.get_unused_code(sort_by_size=True)] == ['bar', 'foo']
+    assert [item.name for item in
+            v.get_unused_code(sort_by_size=False)] == ['foo', 'bar']
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,3 +1,5 @@
+import pytest
+
 from . import v
 assert v  # Silence pyflakes.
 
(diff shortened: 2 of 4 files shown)


tests/test_sorting.py
```diff
@@ -0,0 +1,16 @@
...
+def hello():
+    pass
+
+def world():
+    print("Hello, I am a long function.")
+    return "World"
+""")
+    output = [item.name for item in v.get_unused_code(sort_by_size=True)]
+    expected = ['hello', 'world']
+    assert output == expected
```
jendrikseipp: Let's make hello() longer than world(). Then we can have two tests: `assert [item.name for item in v.get_unused_code(sort_by_size=False)] == ['hello', 'world']` and `assert [item.name for item in v.get_unused_code(sort_by_size=True)] == ['world', 'hello']`. Also, let's use 'foo' and 'bar' instead of 'hello' and 'world'. "Hello world" is generally used for simple programs, not for testing.

tests/test_sorting.py
```diff
@@ -0,0 +1,16 @@
+from . import v
+assert v  # Silence flake 8
```
jendrikseipp: `# Silence pyflakes`

tests/test_sorting.py
RJ722: Agreed!

jendrikseipp:
Thanks!

PR #135, opened 2018-06-30
Add make whitelist

RJ722:
Add ``--make-whitelist`` flag - Report output in whitelist ready format.

## Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -51,6 +51,7 @@ Usage
   $ vulture myscript.py mypackage/
   $ vulture myscript.py mywhitelist.py  # Put false-positives in whitelist.
   $ vulture myscript.py --min-confidence 100  # Only report 100% dead code.
+  $ vulture myscript.py --make-whitelist  # Report in whitelist ready format.
 
 The provided arguments may be Python files or directories. For each
 directory Vulture analyzes all contained `*.py` files.
--- a/README.rst
+++ b/README.rst
@@ -102,6 +102,15 @@ its number of lines. This helps developers prioritize where to look for
 dead code first.
 
 
+Make whitelist
+--------------
+
+When ``--make-whitelist`` is enabled, Vulture reports output in a format such
+that it is ready to serve as a whitelist. Use it when you've ensured that
+the unsed code has already been removed and Vulture is only reporting false
+positives.
+
+
 Examples
 --------
 
--- a/README.rst
+++ b/README.rst
@@ -51,7 +51,7 @@ Usage
   $ vulture myscript.py mypackage/
   $ vulture myscript.py mywhitelist.py  # Put false-positives in whitelist.
   $ vulture myscript.py --min-confidence 100  # Only report 100% dead code.
-  $ vulture myscript.py --make-whitelist  # Report in whitelist ready format.
+  $ vulture myscript.py --make-whitelist  # Report in whitelist format.
 
 The provided arguments may be Python files or directories. For each
 directory Vulture analyzes all contained `*.py` files.
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -25,11 +25,12 @@ def check(items_or_names, expected_names):
     try:
         # items_or_names is a collection of Item objects.
         names = sorted(item.name for item in items_or_names)
-        expected_names = sorted(expected_names)
-        assert names == expected_names
     except AttributeError:
         # items_or_names is a set of strings.
         assert items_or_names == set(expected_names)
+    else:
+        expected_names = sorted(expected_names)
+        assert names == expected_names
 
 
 def check_unreachable(v, lineno, size, name):
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,6 +1,8 @@
 import ast
 import glob
 import os.path
+import subprocess
+import sys
 
 import pytest
 from vulture import core
@@ -14,15 +16,20 @@ skip_if_not_has_async = pytest.mark.skipif(
     reason="needs async support (added in Python 3.5)")
 
 
+def call_vulture(args, **kwargs):
+    return subprocess.call(
+        [sys.executable, '-m', 'vulture'] + args, cwd=REPO, **kwargs)
+
+
 def check(items_or_names, expected_names):
-    if isinstance(items_or_names, set):
-        # items_or_names is a set of strings.
-        assert items_or_names == set(expected_names)
-    else:
-        # items_or_names is a list of Item objects.
+    try:
+        # items_or_names is a collection of Item objects.
         names = sorted(item.name for item in items_or_names)
         expected_names = sorted(expected_names)
         assert names == expected_names
+    except AttributeError:
+        # items_or_names is a set of strings.
+        assert items_or_names == set(expected_names)
 
 
 def check_unreachable(v, lineno, size, name):
--- a/tests/test_make_whitelist.py
+++ b/tests/test_make_whitelist.py
@@ -55,6 +55,15 @@ from lorem import ipsum as dolor
 def test_unused_attribute(check_whitelist):
     code = """\
 class Foo:
+    def bar(self):
+        self.foobar = 'unused attr'
+"""
+    check_whitelist(code, ['Foo', 'bar', 'foobar'], [])
+
+
+def test_unused_property(check_whitelist):
+    code = """\
+class Foo:
     @property
     def bar(self):
         pass
--- a/tests/test_make_whitelist.py
+++ b/tests/test_make_whitelist.py
@@ -59,7 +59,7 @@ class Foo:
     def bar(self):
         pass
 """
-    check_whitelist(code, 1)
+    check_whitelist(code)
 
 
 def test_unreachable_code(check_whitelist):
--- a/tests/test_make_whitelist.py
+++ b/tests/test_make_whitelist.py
@@ -1,0 +1,71 @@
+import pytest
+
+from . import call_vulture, v
+assert v  # silence pyflakes
+
+
+@pytest.fixture
+def check_whitelist(v, tmpdir, capsys):
+    def examine(code, expected_out=0, min_confidence=0, sort_by_size=False):
+        filename = str(tmpdir.join('myscript.py'))
+        whitelist = str(tmpdir.join('whitelist.py'))
+        with open(filename, 'w') as f:
+            f.write(code)
+        v.scavenge([filename])
+        capsys.readouterr().out
+        v.make_whitelist(min_confidence, sort_by_size)
+        with open(whitelist, 'w') as f:
+            f.write(capsys.readouterr().out)
+        assert call_vulture([filename, whitelist]) == expected_out
+    return examine
+
+
+def test_unused_function(check_whitelist):
+    code = """\
+def func():
+    pass
+"""
+    check_whitelist(code)
+
+
+def test_unused_class(check_whitelist):
+    code = """\
+class Foo:
+    def __init__(self):
+        pass
+"""
+    check_whitelist(code)
+
+
+def test_unused_variables(check_whitelist):
+    code = """\
+foo = 'unused'
+bar = 'variable'
+"""
+    check_whitelist(code)
+
+
+def test_unused_import(check_whitelist):
+    code = """\
+import this
+"""
+    check_whitelist(code, 1)
+
+
+def test_unused_attribute(check_whitelist):
+    code = """\
+class Foo:
+    @property
+    def bar(self):
+        pass
+"""
+    check_whitelist(code, 1)
+
+
+def test_unreachable_code(check_whitelist):
+    code = """\
+def foo():
+    return "Foo Bar"
+    print("Hello")
+"""
+    check_whitelist(code, 1)
--- a/tests/test_make_whitelist.py
+++ b/tests/test_make_whitelist.py
@@ -5,7 +5,7 @@ assert v  # silence pyflakes
 
 
 @pytest.fixture
-def check_whitelist(v, tmpdir, capsys):
+def check_whitelist(v, capsys):
     def examine(code, results_before, results_after):
         v.scan(code)
         check(v.get_unused_code(), results_before)
--- a/tests/test_make_whitelist.py
+++ b/tests/test_make_whitelist.py
@@ -1,22 +1,19 @@
 import pytest
 
-from . import call_vulture, v
+from . import check, v
 assert v  # silence pyflakes
 
 
 @pytest.fixture
 def check_whitelist(v, tmpdir, capsys):
-    def examine(code, expected_out=0, min_confidence=0, sort_by_size=False):
-        filename = str(tmpdir.join('myscript.py'))
-        whitelist = str(tmpdir.join('whitelist.py'))
-        with open(filename, 'w') as f:
-            f.write(code)
-        v.scavenge([filename])
-        capsys.readouterr().out
-        v.make_whitelist(min_confidence, sort_by_size)
-        with open(whitelist, 'w') as f:
-            f.write(capsys.readouterr().out)
-        assert call_vulture([filename, whitelist]) == expected_out
+    def examine(code, results_before, results_after):
+        v.scan(code)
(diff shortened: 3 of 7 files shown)


vulture/core.py
```diff
@@ -255,7 +259,8 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
-    def report(self, min_confidence=0, sort_by_size=False):
+    def report(self, min_confidence=0, sort_by_size=False,
+               make_whitelist=False):
```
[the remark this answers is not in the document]
jendrikseipp: We also have to only report a subset of unused items when `--make-whitelist` is active: e.g., unreachable code can't be added to a whitelist. Therefore, I recommend adding a new method `Vulture.make_whitelist()` instead of adding the functionality to `Vulture.report()`.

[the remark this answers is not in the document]
RJ722: Hey @jendrikseipp, I saw your email just now after pushing the commit. Making amendments now.

vulture/core.py
```diff
@@ -255,6 +255,17 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
+    def make_whitelist(self, min_confidence=0, sort_by_size=False):
+        for item in self.get_unused_code(
+                min_confidence=min_confidence, sort_by_size=sort_by_size):
+            self.found_dead_code_or_error = True
```
jendrikseipp: Let's move this line below the continue statement to exit with 0 if there's no output.

vulture/core.py
```diff
@@ -255,6 +255,17 @@ def by_size(item):
...
+            if item.typ in ('unreachable_code', 'import'):
+                continue
+            print("{0} # unused {1} ({2}:{3:d})".format(
```
jendrikseipp: Remove numbers between brackets.

vulture/core.py
```diff
@@ -505,6 +516,9 @@ def csv(exclude):
         ' *settings.py,docs/*.py). PATTERNs can contain globbing characters'
         ' (*, ?, [, ]). Treat PATTERNs without globbing characters as'
         ' *PATTERN*.')
+    parser.add_argument(
+            '--make-whitelist', action='store_true', help='Report unused code'
```
jendrikseipp: Please start `help` in separate line.

vulture/core.py
```diff
@@ -521,6 +535,10 @@ def main():
     args = _parse_args()
     vulture = Vulture(verbose=args.verbose)
     vulture.scavenge(args.paths, exclude=args.exclude)
+    if args.make_whitelist:
```
jendrikseipp: Let's use `report_func = vulture.make_whitelist if args.make_whitelist else vulture.report` and then `sys.exit(report_func(...))` to avoid repeating code.

tests/test_report.py
```diff
@@ -0,0 +1,132 @@
...
+        print("A small unused function")
+"""
+
+
+@pytest.fixture
+def check_report(v, capsys, tmpdir):
+    def test_report(code, expected, min_confidence=0, sort_by_size=False,
+                    make_whitelist=False):
+        filename = str(tmpdir.join('dead_code.py'))
+        with open(filename, 'w') as f:
```
jendrikseipp: Do we really need temporary files to test the new feature?

vulture/core.py
```diff
@@ -255,6 +255,17 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
+    def make_whitelist(self, min_confidence=0, sort_by_size=False):
+        for item in self.get_unused_code(
+                min_confidence=min_confidence, sort_by_size=sort_by_size):
+            self.found_dead_code_or_error = True
+            if item.typ in ('unreachable_code', 'import'):
```
jendrikseipp: Let's explicitly list the types that are supported here instead of listing the ones that aren't.

tests/test_report.py
RJ722: I guess, it makes tests pretty robust since tests will break if any part of the pipeline is tampered with. But, still I'd say it's your call.

jendrikseipp: I forgot that `Vulture.make_whitelist()` directly writes to stdout.

vulture/core.py
```diff
@@ -255,6 +255,17 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
+    def make_whitelist(self, min_confidence=0, sort_by_size=False):
+        for item in self.get_unused_code(
+                min_confidence=min_confidence, sort_by_size=sort_by_size):
+            if item.typ in ('function', 'class', 'property', 'variable',
+                            'attribute'):
+                print("{} # unused {} ({}:{:d})".format(
```
jendrikseipp: Let's leave two spaces before `#`.

RJ722:
The [job on `pypy`](https://travis-ci.org/jendrikseipp/vulture/jobs/398651102) seems to have stalled, can you please restart it?

README.rst
```diff
@@ -51,6 +51,7 @@ Usage
   $ vulture myscript.py mypackage/
   $ vulture myscript.py mywhitelist.py  # Put false-positives in whitelist.
   $ vulture myscript.py --min-confidence 100  # Only report 100% dead code.
+  $ vulture myscript.py --make-whitelist  # Report in whitelist format.
```
jendrikseipp: Can you move this example to a dedicated section that explains how and why to use the new feature, please?

tests/test_report.py
```diff
@@ -0,0 +1,132 @@
+import pytest
```
jendrikseipp: The tests in this file are rather hard to maintain, as you might have seen when adding the extra space before the `#`. I think we have to think about what we actually want to test here. Having one or two tests that check the output is fine, but we already have other tests for checking the sorting and the filtering by minimum confidence.

On the other hand, some tests are missing. For example, I don't think the `--make-whitelist` feature works for properties or attributes at the moment. We should check that `vulture --make-whitelist myfile.py > whitelist.py` results in `vulture whitelist.py myfile.py` exiting with 0.

RJ722: Yes, indeed the previous version of tests was very verbose and fragile - I've only very basic tests for ensuring that the output format remains intact. Also, checking for the exit code was a great idea - I've tried implementing a version of it in `tests/test_make_whitelist.py`. I've also included the missing tests for attributes and properties. Let me know if it could be improved.

vulture/core.py
```diff
@@ -255,6 +255,16 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
+    def make_whitelist(self, min_confidence=0, sort_by_size=False):
+        for item in self.get_unused_code(
+                min_confidence=min_confidence, sort_by_size=sort_by_size):
+            if item.typ in ('function', 'class', 'variable'):
```
jendrikseipp: What about properties and attributes? We need to add output for them as well.

RJ722: Can we handle it in a separate PR, please?

README.rst
RJ722: Done!

vulture/core.py
jendrikseipp: I'd rather handle this in this PR since the functionality is tightly linked. I wouldn't want to see in the second PR that we have to undo some changes from the first PR.

What we need for properties and attributes (e.g., named `foo`) is the following output at the beginning of the output:

```
class VultureMock:
    def __getattr__(self, _):
        pass

_ = VultureMock()
```
and then for each unused property or attribute:
```
_.foo  # unused attribute (...)
```

tests/test_script.py
```diff
@@ -58,3 +53,8 @@ def call_vulture_with_excludes(excludes):
 
     assert call_vulture_with_excludes(['core.py', 'utils.py']) == 1
     assert call_vulture_with_excludes(['core.py', 'utils.py', 'lines.py']) == 0
+
+
+def test_make_whitelist():
+    assert call_vulture(
+            ['vulture/', '--make-whitelist', '--exclude', 'whitelists']) == 1
```
jendrikseipp: Let's add `assert call_vulture(['vulture/', '--make-whitelist']) == 0`

tests/__init__.py
```diff
@@ -14,15 +16,20 @@
...
+
 def check(items_or_names, expected_names):
-    if isinstance(items_or_names, set):
-        # items_or_names is a set of strings.
-        assert items_or_names == set(expected_names)
-    else:
-        # items_or_names is a list of Item objects.
+    try:
+        # items_or_names is a collection of Item objects.
         names = sorted(item.name for item in items_or_names)
```
jendrikseipp: Only put the line that can fail into the try-block, everything else should go into the else-block.

vulture/core.py
RJ722: I was trying to access the parent node and extract it's name from there,  but this turns out to be way easier than it.😅

Thanks for the help! :-)

vulture/core.py
```diff
@@ -255,6 +255,29 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
+    def make_whitelist(self, min_confidence=0, sort_by_size=False):
+        unused_code = self.get_unused_code(
+                min_confidence=min_confidence, sort_by_size=sort_by_size)
+        if any(item.typ in ('attribute', 'property') for item in unused_code):
+            print("""\
+class VultureMock:
+    def __init__(self, _):
```
jendrikseipp: `__init__` must be `__getattr__`. Given that the tests pass nonetheless, we are not testing the right things...

tests/test_make_whitelist.py
```diff
@@ -0,0 +1,71 @@
+import pytest
+
+from . import call_vulture, v
+assert v  # silence pyflakes
+
+
+@pytest.fixture
+def check_whitelist(v, tmpdir, capsys):
+    def examine(code, expected_out=0, min_confidence=0, sort_by_size=False):
```
jendrikseipp: expected_out -> expected_returncode

jendrikseipp: We don't need to test `min_confidence` and `sort_by_size` here.

jendrikseipp: I'd like to make the tests simpler and faster in the following way (untested):

```
def examine(code, results_before, results_after):
    v.scan(code)
    assert v.get_unused_code() == results_before
    capsys.readouterr()  # Clear captured text (might be unnecessary).
    v.make_whitelist()
    whitelist, _ = capsys.readouterr()
    v.scan(whitelist)
    assert v.get_unused_code() == results_after
```

vulture/core.py
RJ722: Yes, that's a mistake. Although this code is erred, Vulture successfully gets the unused attribute/property into the `used_attr` set and therefore starts whitelisting the results. So, I guess we might be testing the correct thing, am I correct?

RJ722: We might want to check the sanctity of the code by running it with python, but I'm not sure that we would want to do that, should we?

tests/test_make_whitelist.py
RJ722: Yes, that works out of the box! ;-)

vulture/core.py
jendrikseipp: The resulting whitelist must be syntactically correct, but it usually can't be executed. Our tests chek the syntax in the scan method, so that's fine. Since the result can't be excuted, we also don't need the VultureMock class after all. We are testing the correct thing after all :-)

vulture/core.py
```diff
@@ -255,6 +255,29 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
+    def make_whitelist(self, min_confidence=0, sort_by_size=False):
+        unused_code = self.get_unused_code(
+                min_confidence=min_confidence, sort_by_size=sort_by_size)
+        if any(item.typ in ('attribute', 'property') for item in unused_code):
```
jendrikseipp: Could you please use lists instead of tuples for homogeneous collections (here and below)? I.e., square brackets instead of normal brackets.

vulture/core.py
RJ722: So, from what I get - we just print `_.foobar  # unused attrivute (...)` without explicitly declaring what is `_` since that's enough to whitelist foobar from being reported by Vulture, right?

vulture/core.py
RJ722: I am sorry, I'm a little confused 😅. I thought that we were gonna [remove this part](https://github.com/jendrikseipp/vulture/pull/135#discussion_r199561968) because we didn't need `VultureMock` class?

vulture/core.py
jendrikseipp: Right.

vulture/core.py
jendrikseipp: Right, I should have placed the comment somewhere else, e.g., I prefer `['function', 'class', 'variable', 'import']` instead of `('function', 'class', 'variable', 'import')`, etc.

tests/test_report.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @property
+    def myprop(self):
+        pass
+"""
+
+
+@pytest.fixture
+def check_report(v, capsys, tmpdir):
+    def test_report(code, expected, min_confidence=0, sort_by_size=False,
```
jendrikseipp: We don't need to test the sorting here. Also, we don't need temporary files, if we use scan instead of scavenge.

vulture/core.py
RJ722: Changes done, Thanks for the clarification. :-)

tests/test_report.py
RJ722: I wrote this test for ensuring that `size_report` stays intact, but I guess it is better off without at test 😛

RJ722:
I rewrote the last commit, so the review is gone. I've renamed the filename to `foo.py` and removed the comment.

tests/test_report.py
```diff
@@ -0,0 +1,64 @@
...
+    This is a long class
+    '''
+    def __init__(self):
+        print("Initialized foo")
+
+    def bar(self):
+        self.foobar = "unused attribute"
+        foobar = "unused variable"
+        return
+        print("A small unused function")
```
jendrikseipp: This is not a function. Let's just use `print("unreachable")`.

vulture/core.py
```diff
@@ -255,6 +255,21 @@ def by_size(item):
         return sorted(confidently_unused,
                       key=by_size if sort_by_size else by_name)
 
+    def make_whitelist(self, min_confidence=0, sort_by_size=False):
```
jendrikseipp: Let's make this a little simpler:

```
for item in self.get_unused_code(
        min_confidence=min_confidence, sort_by_size=sort_by_size):
    if item.typ != 'unreachable_code':
        prefix = '_.' if item.typ in ['attribute', 'property'] else ''
        print("{}{}  # unused {} ({}:{:d})".format(
            prefix, item.name, item.typ, utils.format_path(item.filename),
            item.first_lineno))
return self.found_dead_code_or_error
```

tests/test_report.py
jendrikseipp: Let's use `print("unreachable")` instead.

tests/test_report.py
```diff
@@ -0,0 +1,64 @@
...
+from . import v
+assert v  # Silence pyflakes
+
+mock_code = """\
+import foo
+
+class Foo:
+    '''
+    This is a long class
+    '''
```
jendrikseipp: Please remove the docstring.

vulture/core.py
RJ722: Thanks! :-)

tests/test_report.py
RJ722: oops! :-p

jendrikseipp:
Thanks, Rahul!

RJ722:
You're most welcome. It was a tough one, couldn't have pulled it without you, thanks! :-)

Issue #136, opened 2018-06-30
Remove (most) mock objects from whitelist files


jendrikseipp:
We currently mock all objects in our whitelist files, e.g. `whitelist_defaultdict = Whitelist(); whitelist_defaultdict.default_factory`. This has the drawback that we can't test whether `default_factory` really exists (and is not misspelled or missing from the stdlib in this Python version). For non-dynamic objects, we should therefore access the objects directly instead of their mocks. The only place where we have to keep mock objects is the whitelist for the `ast` module, since these methods are created dynamically. To allow importing `collections` in the `collections.py` whitelist file, we have to rename it to `collections_whitelist.py` to avoid name clashes. Here is a diff that shows the necessary changes for `collections.py`:

```
diff --git a/vulture/core.py b/vulture/core.py
index 5400ae6..4f43e80 100644
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -218,7 +218,7 @@ class Vulture(ast.NodeVisitor):
 
         unique_imports = set(item.name for item in self.defined_imports)
         for import_name in unique_imports:
-            path = os.path.join('whitelists', import_name) + '.py'
+            path = os.path.join('whitelists', import_name) + '_whitelist.py'
             if exclude_file(path):
                 self._log('Excluded whitelist:', path)
             else:
diff --git a/vulture/whitelists/collections_whitelist.py b/vulture/whitelists/collections_whitelist.py
index 4af0cdd..c43bc58 100644
--- a/vulture/whitelists/collections_whitelist.py
+++ b/vulture/whitelists/collections_whitelist.py
@@ -1,5 +1,4 @@
-from whitelist_utils import Whitelist
+import collections
 
 # To free memory, the "default_factory" attribute can be set to None.
-whitelist_defaultdict = Whitelist()
-whitelist_defaultdict.default_factory
+collections.defaultdict.default_factory
```

PR #137, opened 2018-06-30
Refer non-dynamic objects directly instead of their mocks.

RJ722:
Closes: #136

## Checklist:


NEWS.rst
```diff
@@ -6,6 +6,7 @@ News
 * Don't import any Vulture modules in setup.py (thanks @RJ722).
 * Add ``*-test.py`` to recognized test file patterns.
 * Add ``failureException``, ``longMessage`` and ``maxDiff`` to ``unittest`` whitelist.
+* Refer to actual objects rather than their mockups in default whitelists.
```
jendrikseipp: Please update Todo file as well by removing the corresponding item.

vulture/core.py
```diff
@@ -218,7 +218,7 @@ def exclude_file(name):
 
         unique_imports = set(item.name for item in self.defined_imports)
         for import_name in unique_imports:
-            path = os.path.join('whitelists', import_name) + '.py'
+            path = os.path.join('whitelists', import_name) + '_whitelist.py'
```
jendrikseipp: The argparse file should be changed as well.

vulture/whitelists/unittest_whitelist.py
```diff
@@ -0,0 +1,22 @@
...
+unittest.TestCase.tearDown
+unittest.TestCase.setUpClass
+unittest.TestCase.tearDownClass
+unittest.TestCase.run
+unittest.TestCase.skipTest
+unittest.TestCase.debug
+unittest.TestCase.failureException
+unittest.TestCase.longMessage
+unittest.TestCase.maxDiff
+if hasattr(unittest.TestCase, 'subTest'):
```
jendrikseipp: Let's use a try/except here to be less repetitive.

vulture/whitelists/unittest_whitelist.py
```diff
@@ -0,0 +1,22 @@
...
+    # new in Python 3.4
+    unittest.TestCase.subTest
+
+# unittest.mock
+whitelist_mock = Whitelist()
+whitelist_mock.return_value
+whitelist_mock.side_effect
```
jendrikseipp: Can we remove the need for `Whitelist` here as well?

vulture/core.py
RJ722: Actually, I was wondering if we can directly access `epilog` since it's just an argument to the `ArgumentParser` class. Is it possible without actually instantiating the class?

jendrikseipp: If not, we can just instantiate it.

RJ722: Hey @jendrikseipp,

I just conducted a little experiment:

```
In [1]: from vulture import Vulture

In [2]: v = Vulture()

In [3]: v.scan("""\
   ...: import argparse
   ...:
   ...: parser = argparse.ArgumentParser(epilog=True)
   ...: """)

In [4]: v.get_unused_code()
Out[4]: ['parser']
```

Why do we want to whitelist `epilog` when Vulture isn't reporting as unused? Did I miss something (which is very probable given how sharp I am :D)?

jendrikseipp: The epliog variable here will be reported as unsued: https://bitbucket.org/jendrikseipp/lab/src/d459dc7d09b120d9c156a866f574101216bb4cd4/lab/experiment.py#lines-38

vulture/whitelists/argparse_whitelist.py
```diff
@@ -0,0 +1,7 @@
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.epilog
```
jendrikseipp: Let's rename `parser` to `_`. Otherwise, Vulture will miss unused variables called `parser`.

RJ722: Oh yes, nice catch!

vulture/whitelists/unittest_whitelist.py
```diff
@@ -0,0 +1,25 @@
...
+try:
+    # new in Python 3.4
+    TestCase.subTest
+except AttributeError:
+    pass
+
+try:
+    from unittest import mock
+    # unittest.mock was introduced in Python 3.3
+    mock.Mock.return_value
```
jendrikseipp: Let's move the two actual usages to an else block.

NEWS.rst
jendrikseipp: mockups -> mocks

jendrikseipp:
Thanks!

PR #139, opened 2018-07-04
Add --ignore-names flag

RJ722:
## Description
Ignore a certain regular expression across the results reported by Vulture.

## Checklist:

- [x] I have added an entry in NEWS.rst.

### DIFF
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -3,7 +3,7 @@ News
 
 0.29 (unreleased)
 -----------------
-* Add ``--ignore-names`` flag for ignoring names matching the given glob pattern.
+* Add ``--ignore-names`` flag for ignoring names matching the given glob patterns.
 
 
 0.28 (2018-07-05)
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -1,6 +1,11 @@
 News
 ====
 
+0.29 (unreleased)
+-----------------
+* Add ``--ignore-names`` flag for ignoring names matching the given glob pattern.
+
+
 0.28 (2018-07-05)
 -----------------
 * Add ``--make-whitelist`` flag for reporting output in whitelist format (thanks @RJ722).
--- a/README.rst
+++ b/README.rst
@@ -71,12 +71,19 @@ We collect whitelists for common Python modules and packages in
 ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 ``--exclude *settings.py,docs/``).
 
-**Ignoring certain results**
+**Ignoring names**
 
-You can use the ``--ignore-names`` option to prevent Vulture from reporting
-item whose names follow a certain regex (for eg. ``[abc]``). Please note
-that for functions, the names of decorators are also matched against the
-given regular expression(s).
+You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
+starting with foo and the names ``bar`` and ``baz``. The ``--ignore-names``
+option also checks decorator names. This is helpful for example in Flask
+projects, where you can use ``--ignore-names app.route`` to ignore all
+functions with the ``@app.route`` decorator.
+
+We recommend using whitelists instead of ``--ignore-names`` whenever
+possible, since whitelists are automatically checked for syntactic
+correctness when passed to Vulture and sometimes you can even pass
+them to your Python interpreter and let it check that all whitelisted
+code actually still exists in your project.
 
 **Marking unused variables**
 
--- a/README.rst
+++ b/README.rst
@@ -71,6 +71,13 @@ We collect whitelists for common Python modules and packages in
 ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 ``--exclude *settings.py,docs/``).
 
+**Ignoring certain results**
+
+You can use the ``--ignore-names`` option to prevent Vulture from reporting
+item whose names follow a certain regex (for eg. ``[abc]``). Please note
+that for functions, the names of decorators are also matched against the
+given regular expression(s).
+
 **Marking unused variables**
 
 There are situations where you can't just remove unused variables, e.g.,
--- a/README.rst
+++ b/README.rst
@@ -74,10 +74,7 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 **Ignoring names**
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with foo and the names ``bar`` and ``baz``. The ``--ignore-names``
-option also checks decorator names. This is helpful for example in Flask
-projects, where you can use ``--ignore-names app.route`` to ignore all
-functions with the ``@app.route`` decorator.
+starting with foo and the names ``bar`` and ``baz``.
 
 We recommend using whitelists instead of ``--ignore-names`` whenever
 possible, since whitelists are automatically checked for syntactic
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 **Ignoring names**
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with foo and the names ``bar`` and ``baz``.
+starting with ``foo`` and the names ``bar`` and ``baz``.
 
 We recommend using whitelists instead of ``--ignore-names`` whenever
 possible, since whitelists are automatically checked for syntactic
--- a/tests/test_ignore_names.py
+++ b/tests/test_ignore_names.py
@@ -16,12 +16,14 @@ def check_ignore():
 
 def test_var(check_ignore):
     code = """\
-foo = 1
+fio = 1
+fao = 2
 bar = 2
-three = 3
+ftobar = 3
+baz = 10000
 funny = True
 """
-    check_ignore(code, ['f*', 'bar'], ['three'])
+    check_ignore(code, ['f?o*', 'ba[rz]'], ['funny'])
 
 
 def test_function(check_ignore):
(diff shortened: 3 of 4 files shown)


vulture/core.py
```diff
@@ -276,6 +285,9 @@ def report(self, min_confidence=0, sort_by_size=False):
             self.found_dead_code_or_error = True
         return self.found_dead_code_or_error
 
+    def ignored(self, name):
```
jendrikseipp: ignored -> _ignore_name

vulture/core.py
```diff
@@ -143,7 +145,12 @@ def __hash__(self):
 class Vulture(ast.NodeVisitor):
     """Find dead code."""
 
-    def __init__(self, verbose=False):
+    def __init__(self, verbose=False, ignore_names=[]):
```
jendrikseipp: [] -> None (http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments)

and below:

`ignore_names = ignore_names or []`

vulture/core.py
```diff
@@ -374,22 +386,41 @@ def visit_AsyncFunctionDef(self, node):
         return self.visit_FunctionDef(node)
 
     def visit_Attribute(self, node):
+        if self.ignored(node.attr):
+            self._log('Ignoring Attribute "{}" (ignored name)'.format(
+                node.attr))
+            return
```
jendrikseipp: Can we make this check once for all types in `Vulture._define` and fold it into the ignore check that's already there?

vulture/core.py
```diff
@@ -374,22 +386,41 @@ def visit_AsyncFunctionDef(self, node):
...
+        if self.ignored(node.name):
+            self._log('Ignoring class "{}" (ignored name)'.format(node.name))
+            return
         self._define(
             self.defined_classes, node.name, node, ignore=_ignore_class)
 
     def visit_FunctionDef(self, node):
         for decorator in node.decorator_list:
-            if getattr(decorator, 'id', None) == 'property':
+            n = decorator.func if isinstance(
```
jendrikseipp: One letter variable names are usually too short.

vulture/core.py
```diff
@@ -143,7 +145,12 @@ def __hash__(self):
 class Vulture(ast.NodeVisitor):
     """Find dead code."""
 
-    def __init__(self, verbose=False):
+    def __init__(self, verbose=False, ignore_names=[]):
+        for name in ignore_names:
+            if any(fnmatchcase(unsafe_name, name)
+                    for unsafe_name in UNSAFE_IGNORE_NAMES):
+                raise ValueError(
+                        'It is not safe to ignore the name "{}".'.format(name))
```
jendrikseipp: I think it would be better if we could just ignore the `ignore_names` list when checking for the string "property". We probably habe to test for "property" before deciding whether we cant to ignore the property.

vulture/core.py
```diff
@@ -384,7 +389,11 @@ def visit_ClassDef(self, node):
 
     def visit_FunctionDef(self, node):
         for decorator in node.decorator_list:
-            if getattr(decorator, 'id', None) == 'property':
+            decorator = decorator.func if isinstance(
+                decorator, ast.Call) else decorator
+            name = decorator.attr if isinstance(
+                decorator, ast.Attribute) else decorator.id
```
jendrikseipp: These two if statements give rise to 4 different code paths, right? Can we add tests for all 4 of them?

jendrikseipp: It would be helpful if we could use standard if-else statements here to see that all lines are tested.

NEWS.rst
```diff
@@ -1,6 +1,11 @@
 News
 ====
 
+0.29 (unreleased)
+-----------------
+* Add ``--ignore-names`` flag for ignoring names matching the given glob pattern.
```
jendrikseipp: pattern -> patterns

README.rst
```diff
@@ -71,6 +71,13 @@ We collect whitelists for common Python modules and packages in
 ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 ``--exclude *settings.py,docs/``).
 
+**Ignoring certain results**
+
+You can use the ``--ignore-names`` option to prevent Vulture from reporting
+item whose names follow a certain regex (for eg. ``[abc]``). Please note
+that for functions, the names of decorators are also matched against the
+given regular expression(s).
```
jendrikseipp: `**Ignoring names**`

You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names starting with ``foo`` and the names ``bar`` and ``baz``. The ``--ignore-names`` option also checks decorator names. This is helpful for example in Flask projects, where you can use ``--ignore-names app.route`` to ignore all functions with the ``@app.route`` decorator.

We recommend using whitelists instead of ``--ignore-names`` whenever possible, since whitelists are automatically checked for syntactic correctness when passed to Vulture and sometimes you can even pass them to your Python interpreter and let it check that all whitelisted code actually still exists in your project.

tests/test_ignore_names.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_attribute(check_ignore):
+    code = """\
+class Foo:
+    def __init__(self, attr_foo, attr_bar):
+        self.attr_foo = attr_foo
+        self.attr_bar = attr_bar
+"""
+    check_ignore(code, ['*_foo'], ['Foo', 'attr_bar'])
```
jendrikseipp: We also need tests for decorator names and imports and tests for ``ba[rz]`` and ``f?o``. Also, we need to check that the matching is case-sensitive (ignoring foo doesn't ignore Foo).

vulture/core.py
```diff
@@ -275,6 +277,9 @@ def report(self, min_confidence=0, sort_by_size=False):
             self.found_dead_code_or_error = True
         return self.found_dead_code_or_error
 
+    def _ignore_name(self, name):
```
jendrikseipp: This can become a local function in `_define`.

vulture/core.py
RJ722: yes, that would be helpful coverage wise. Also, it helped me uncover a foolish bug -  that I forgot to ignoring the function in this branch. 😅 

Fix arriving.

vulture/core.py
RJ722: Since, we don't use `_define` for decorators, I think we would need to keep it separate, so that it can still be used to ignore decorated functions.

jendrikseipp: You're right.

vulture/core.py
jendrikseipp: It seems that the if statements always evaluate to True: https://coveralls.io/builds/18099744/source?filename=vulture/core.py (6 out of 6 test cases). Can you please add tests where the if-statement is false? Or does this never happen and we can remove the if?

RJ722: I am not quite sure how this is happening because _at least_ the `@property` case dissatisfies both the conditions.
I also tested the other test cases locally - both the False conditions are tested:

```python
In [1]: import vulture; v = vulture.Vulture()

In [2]: code = """\
   ...: @property
   ...: def rahul():
   ...:     pass
   ...: """

In [3]: v.scan(code)
decorator.id

In [4]: code = """\
   ...: @f.foobar
   ...: def rahil():
   ...:     pass
   ...: """

In [5]: v.scan(code)
Ast.Attribute

In [6]: code = """\
   ...: @f.foobar()
   ...: def rahul():
   ...:     pass
   ...: """

In [7]: v.scan(code)
Ast.call
Ast.Attribute
```

I don't know how, but it also displays lines under both `if` and `else` 6x simultaneously. 😅

RJ722: In the above experiment, I've just added `print` statements with branch names:
```python
if isinstance(decorator, ast.Call):
    decorator = decorator.func
    print("Ast.call")
if isinstance(decorator, ast.Attribute):
    name = decorator.attr
    print("Ast.Attribute")
else:
    name = decorator.id
    print("decorator.id")

```

jendrikseipp: OK, then I guess we shouldn't care too much about the counts in coveralls.io. It would be great to know if codecov has the same problems or whether it reports the executions more accurately. Could you open an issue for this, please?

RJ722: Sure, here: https://github.com/jendrikseipp/vulture/issues/141

vulture/core.py
```diff
@@ -384,9 +389,20 @@ def visit_ClassDef(self, node):
 
     def visit_FunctionDef(self, node):
         for decorator in node.decorator_list:
-            if getattr(decorator, 'id', None) == 'property':
+            if isinstance(decorator, ast.Call):
+                decorator = decorator.func
+            if isinstance(decorator, ast.Attribute):
+                name = decorator.attr
+            else:
+                name = decorator.id
```
jendrikseipp: This looks like we can't actually ignore "app.route", right? This contradicts the documentation. I think we should take the full decorator "name" into account, e.g., `f.foobar()` can be ignored with `f.foobar()` but not with `foobar`. I can imagine that we need a helper function for retrieving the decorator name from a decorator node, since these names can be nested more deeply, I guess, like `a.b.c.foo`.

Also, I think for ignoring decorators we should require "@" signs. So the documentation should talk about `--ignore-names "@app.route*"` (note the quotes which prevent bash from expanding the input).

vulture/core.py
```diff
@@ -515,6 +531,11 @@ def csv(exclude):
         ' *settings.py,docs/*.py). PATTERNs can contain globbing characters'
         ' (*, ?, [, ]). Treat PATTERNs without globbing characters as'
         ' *PATTERN*.')
+    parser.add_argument(
+            '--ignore-names', metavar='PATTERN', type=csv, default=None,
+            help='Comma-seperated list of names to ignore (e.g.,'
+            ' app.route). Arguments may contain globbing characters (*, ?,'
```
jendrikseipp: seperated -> separated

jendrikseipp: app.route -> `visit_*,@app.route*`

vulture/core.py
jendrikseipp: Actually, I think we should ignore the round brackets and everything in them. So `@app.route` ignores the Flask decorators.

RJ722: Yes, you are completely right. I don't how this slipped by. 😕. Let me make the changes.

vulture/core.py
jendrikseipp: This still needs fixing.

README.rst
```diff
@@ -71,6 +71,17 @@ We collect whitelists for common Python modules and packages in
 ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 ``--exclude *settings.py,docs/``).
 
+**Ignoring names**
+
+You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
+starting with foo and the names ``bar`` and ``baz``.
```
jendrikseipp: foo -> ``` ``foo`` ```

RJ722:
I really should add a spell checker. 😓

jendrikseipp:
Thanks, Rahul!

PR #142, opened 2018-07-24
Compare functions' decorator against ignore names

RJ722:
Currently, the ``--ignore-names`` flag only ignores names of entities whose names matches with the given input. After this PR is merged, users would be able to ignore functions on the basis of their decorators. For example, `vulture myproject.py --ignore-names "@app.route"` ignores all the functions which are decorated with `app.route`.

## Checklist:

- [x] Add tests for `get_decorator_name`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,17 +74,17 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 **Ignoring names**
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with ``foo`` and the names ``bar`` and ``baz``. The ``--ignore-names``
-option if started with a ``@`` symbol checks for functions decorated with the
-given decorator. This is helpful for example in Flask projects, where you can
-use ``--ignore-names "@app.route"`` to ignore all functions with the
-``@app.route`` decorator.
+starting with ``foo`` and the names ``bar`` and ``baz``. Additionally, the
+``--ignore-decorators`` option should be used to ignore functions decorated
+with the given decorator. This is helpful for example in Flask projects,
+where you can use ``--ignore-names app.route`` to ignore all functions
+with the ``app.route`` decorator.
 
-We recommend using whitelists instead of ``--ignore-names`` whenever
-possible, since whitelists are automatically checked for syntactic
-correctness when passed to Vulture and sometimes you can even pass
-them to your Python interpreter and let it check that all whitelisted
-code actually still exists in your project.
+We recommend using whitelists instead of ``--ignore-names`` or
+``--ignore-decorator`` whenever possible, since whitelists are automatically
+checked for syntactic correctness when passed to Vulture and sometimes
+you can even pass them to your Python interpreter and let it check that
+all whitelisted code actually still exists in your project.
 
 **Marking unused variables**
 
--- a/README.rst
+++ b/README.rst
@@ -75,16 +75,16 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
 starting with ``foo`` and the names ``bar`` and ``baz``. Additionally, the
-``--ignore-decorators`` option should be used to ignore functions decorated
+``--ignore-decorators`` option can be used to ignore functions decorated
 with the given decorator. This is helpful for example in Flask projects,
-where you can use ``--ignore-names "@app.route"`` to ignore all functions
-with the ``app.route`` decorator.
+where you can use ``--ignore-decorators "@app.route"`` to ignore all functions
+with the ``@app.route`` decorator.
 
 We recommend using whitelists instead of ``--ignore-names`` or
-``--ignore-decorator`` whenever possible, since whitelists are automatically
-checked for syntactic correctness when passed to Vulture and sometimes
-you can even pass them to your Python interpreter and let it check that
-all whitelisted code actually still exists in your project.
+``--ignore-decorators`` whenever possible, since whitelists are automatically
+checked for syntactic correctness when passed to Vulture and often you can
+even pass them to your Python interpreter and let it check that all
+whitelisted code actually still exists in your project.
 
 **Marking unused variables**
 
--- a/README.rst
+++ b/README.rst
@@ -75,9 +75,10 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
 starting with ``foo`` and the names ``bar`` and ``baz``. The ``--ignore-names``
-option also checks decorator names. This is helpful for example in Flask
-projects, where you can use ``--ignore-names app.route`` to ignore all
-functions with the ``@app.route`` decorator.
+option if started with a ``@`` symbol checks for functions decorated with the
+given decorator. This is helpful for example in Flask projects, where you can
+use ``--ignore-names "@app.route"`` to ignore all functions with the
+``@app.route`` decorator.
 
 We recommend using whitelists instead of ``--ignore-names`` whenever
 possible, since whitelists are automatically checked for syntactic
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,10 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 **Ignoring names**
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with ``foo`` and the names ``bar`` and ``baz``.
+starting with foo and the names ``bar`` and ``baz``. The ``--ignore-names``
+option also checks decorator names. This is helpful for example in Flask
+projects, where you can use ``--ignore-names app.route`` to ignore all
+functions with the ``@app.route`` decorator.
 
 We recommend using whitelists instead of ``--ignore-names`` whenever
 possible, since whitelists are automatically checked for syntactic
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,10 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 **Ignoring names**
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with ``foo`` and the names ``bar`` and ``baz``.
+starting with ``foo`` and the names ``bar`` and ``baz``. The ``--ignore-names``
+option also checks decorator names. This is helpful for example in Flask
+projects, where you can use ``--ignore-names app.route`` to ignore all
+functions with the ``@app.route`` decorator.
 
 We recommend using whitelists instead of ``--ignore-names`` whenever
 possible, since whitelists are automatically checked for syntactic
--- a/README.rst
+++ b/README.rst
@@ -77,7 +77,7 @@ You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
 starting with ``foo`` and the names ``bar`` and ``baz``. Additionally, the
 ``--ignore-decorators`` option should be used to ignore functions decorated
 with the given decorator. This is helpful for example in Flask projects,
-where you can use ``--ignore-names app.route`` to ignore all functions
+where you can use ``--ignore-names "@app.route"`` to ignore all functions
 with the ``app.route`` decorator.
 
 We recommend using whitelists instead of ``--ignore-names`` or
--- a/README.rst
+++ b/README.rst
@@ -74,10 +74,7 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 **Ignoring names**
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with foo and the names ``bar`` and ``baz``. The ``--ignore-names``
-option also checks decorator names. This is helpful for example in Flask
-projects, where you can use ``--ignore-names app.route`` to ignore all
-functions with the ``@app.route`` decorator.
+starting with foo and the names ``bar`` and ``baz``.
 
 We recommend using whitelists instead of ``--ignore-names`` whenever
 possible, since whitelists are automatically checked for syntactic
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
 **Ignoring names**
 
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with foo and the names ``bar`` and ``baz``.
+starting with ``foo`` and the names ``bar`` and ``baz``.
 
 We recommend using whitelists instead of ``--ignore-names`` whenever
 possible, since whitelists are automatically checked for syntactic
--- a/tests/test_ignore.py
+++ b/tests/test_ignore.py
@@ -98,8 +98,11 @@ def bar():
 def foo():
     pass
 
+@bar
+@foo
 @f.foobar()
 def barfoo():
     pass
 """
     check_ignore(code, ['prop_one'], [], ['decor', 'f.foobar'])
+    check_ignore(code, ['prop_one'], [], ['@decor', '@f.foobar'])
--- a/tests/test_ignore.py
+++ b/tests/test_ignore.py
@@ -3,7 +3,7 @@ from vulture import core
 from . import check, skip_if_not_has_async
 
 
-def check_ignore(code, expected, ignore_names, ignore_decorators):
+def check_ignore(code, ignore_names, ignore_decorators, expected):
     v = core.Vulture(
             verbose=True, ignore_names=ignore_names,
             ignore_decorators=ignore_decorators)
@@ -20,7 +20,7 @@ ftobar = 3
 baz = 10000
 funny = True
 """
-    check_ignore(code, ['funny'], ['f?o*', 'ba[rz]'], [])
+    check_ignore(code, ['f?o*', 'ba[rz]'], [], ['funny'])
 
 
 def test_function():
@@ -34,7 +34,7 @@ def foo():
 def bar():
     pass
 """
-    check_ignore(code, ['bar'], ['foo*'], [])
+    check_ignore(code, ['foo*'], [], ['bar'])
 
 
 @skip_if_not_has_async
@@ -45,7 +45,7 @@ async def foobar():
 async def bar():
     pass
 """
-    check_ignore(code, ['bar'], ['foo*'], [])
+    check_ignore(code, ['foo*'], [], ['bar'])
 
 
 def test_class():
@@ -54,7 +54,7 @@ class Foo:
     def __init__(self):
         pass
 """
-    check_ignore(code, [], ['Foo'], [])
+    check_ignore(code, ['Foo'], [], [])
 
 
 def test_property():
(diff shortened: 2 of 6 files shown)


The-Compiler:
My $0.02: I'm not sure if this should be part of the `--ignore-names` option, for two reasons:

- Discoverability - I was aware of the `--ignore-names` PR and when talking to @RJ722 suggested the idea with ignoring decorated functions. Then he told me that's what the `--ignore-names` PR does as well, which seems kind of unintuitive.
- Potential clashes - if I do `--ignore-names app.route`, both `@app.route` decorated functions as well as `app.route` itself is ignored, right? Also, the other way around. This seems kind of unexpected, because you usually want one or the other, but not both.

So IMHO, this should really be a separate `--ignore-decorators app.route` option (or `--ignore-decorated` or whatever).

RJ722:
> Potential clashes - if I do --ignore-names app.route, both @app.route decorated functions as well as app.route itself is ignored, right? Also, the other way around. This seems kind of unexpected, because you usually want one or the other, but not both.

We add an `@` symbol for adding distinguishing between ignoring decorators and the other stuff. e.g. for ignoring the functions decorated with `app.route`, the command would be:
```
vulture myproject.py --ignore-names "@app.route"
```
whereas, for ignoring a variable (say `app.route`), I'd just type in:
```
vulture myproject.py --ignore-names app.route
```

jendrikseipp:
I still wonder whether a separate `--ignore-decorated` option would be better or worse than having only `--ignore-names`. @The-Compiler and @RJ722 , what are your opinions?

The-Compiler:
Hmm, requiring the `@` solves the clashes, though the documentation is wrong then ("where you can use `--ignore-names app.route` to ignore all functions with the `@app.route` decorator.").

I still have a slight preference for `--ignore-decorated` because of discoverability, but no strong feelings either way.

RJ722:
> Hmm, requiring the @ solves the clashes, though the documentation is wrong then ("where you can use --ignore-names app.route to ignore all functions with the @app.route decorator.").

oops, I forgot to push. :p.

I think that the `--ignore-decorators` was what got us started to work on this issue on the first hand, so let's just keep it as is and make the documentation more clear - which reminds me that I've changed wording a little bit, are there any more recommendations??

jendrikseipp:
I like Florian's discoverability argument, so let's go with `--ignore-decorated`.

jendrikseipp:
I'm undecided on whether we should require `--ignore-decorated @app.route` or `--ignore-decorated app.route`. Or should we just strip leading @ signs? What's your opinion, guys?

RJ722:
IMHO It isn't much helpful if we are going to have ``--ignore-decorators``. Other than that, it also forces people to use quotes because the shell might try to expand the input otherwise.

tests/test_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+@x.y.z
+def doo():
+    pass
+
+@a.b.c.d.foo("Foo and Bar")
+def hoofoo():
+    pass
+"""
+    check_get_decorater_name(code, ['foobar', 'xyz', 'x.y.z', 'a.b.c.d.foo'])
```
jendrikseipp: Can you split this into four tests, please?

vulture/utils.py
```diff
@@ -56,6 +56,18 @@ def format_path(path):
...
 
+def get_decorator_name(decorator):
+    name = ''
+    if isinstance(decorator, ast.Call):
+        decorator = decorator.func
+    while isinstance(decorator, ast.Attribute):
+        name = decorator.attr + ('.' + name if name else '')
+        decorator = decorator.value
+    else:
+        name = decorator.id + ('.' + name if name else '')
```
jendrikseipp: This statement will always be executed, right, because the "while" loop has no "break"? Then we should omit "else:".

vulture/utils.py
```diff
@@ -56,6 +56,18 @@ def format_path(path):
     return relpath if not relpath.startswith('..') else path
 
 
+def get_decorator_name(decorator):
+    name = ''
```
jendrikseipp: Strings should be concatenated with `str.join`. Use `parts = []` to collect the parts and then `return '.'.join(parts)`.

vulture/core.py
```diff
@@ -388,10 +393,18 @@ def visit_ClassDef(self, node):
             self.defined_classes, node.name, node, ignore=_ignore_class)
 
     def visit_FunctionDef(self, node):
```
jendrikseipp: We also need to support async functions and classes.

vulture/core.py
```diff
@@ -351,9 +354,11 @@ def _handle_conditional_node(self, node, name):
 
     def _define(self, collection, name, first_node, last_node=None,
                 message='', confidence=DEFAULT_CONFIDENCE, ignore=None):
+        def _ignore_name(name):
+            return _match(name, self.ignore_names)
         last_node = last_node or first_node
         typ = collection.typ
-        if (ignore and ignore(self.filename, name)) or self._ignore_name(name):
+        if (ignore and ignore(self.filename, name)) or _ignore_name(name):
```
jendrikseipp: `_ignore_name(name)` -> `_match(name, self.ignore_names)`

We also don't need `_ignore_decorator()`.

The-Compiler:
I don't think `@` is special in any shell. Seems fine to just strip it though - I can see people using both, and I don't think there's ever a need to tell the two apart.

jendrikseipp:
I agree, let's strip the leading @'s.

tests/test_utils.py
RJ722: Sure, done! :-)

vulture/utils.py
RJ722: Oh, yes!

vulture/core.py
RJ722: Since we are not discriminating between `async` functions and normal functions inside of Vulture and since both of them share `node.decorator_list`, therefore, I think that we already support them. In any case, I've also added a test case for an async function.

Also, I don't exactly get what do we mean by async classes? 😅

jendrikseipp: I mean regular classes:

```
@foo
class Bar:
    pass
```

vulture/utils.py
```diff
@@ -56,6 +56,18 @@ def format_path(path):
     return relpath if not relpath.startswith('..') else path
 
 
+def get_decorator_name(decorator):
+    parts = []
```
jendrikseipp: move above while

vulture/utils.py
```diff
@@ -56,6 +56,18 @@ def format_path(path):
...
+    if isinstance(decorator, ast.Call):
+        decorator = decorator.func
+    while isinstance(decorator, ast.Attribute):
+        parts.append(decorator.attr)
+        decorator = decorator.value
+    parts.append(decorator.id)
+    parts.reverse()
+    return ".".join(parts)
```
jendrikseipp: `return '.'.join(reversed(parts))` (Note single quotes)

vulture/core.py
```diff
@@ -520,6 +530,11 @@ def csv(exclude):
         help='Comma-separated list of paths to ignore (e.g.,'
         ' "*settings.py,docs/*.py"). {glob_help} A PATTERN without globbing'
         ' characters is treated as *PATTERN*.'.format(**locals()))
+    parser.add_argument(
+        '--ignore-decorators', metavar='PATTERNS', type=csv,
+        help='Comma-separated list of decorators - The functions using these'
+        ' decorators are ignored (e.g., "app.route").'
```
jendrikseipp: Comma-separated list of decorator names. Functions and classes using these decorators are ignored (e.g., "@app.route").

vulture/core.py
```diff
@@ -163,6 +168,7 @@ def get_set(typ):
         self.used_names = get_set('name')
 
         self.ignore_names = ignore_names or []
+        self.ignore_decorators = ignore_decorators or []
```
jendrikseipp: Strip leading @'s in next line and add tests for "@app.route" and "app.route".

tests/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+        node_visitor = ast.NodeVisitor()
+        node_visitor.visit_FunctionDef = visit_FunctionDef
+        node_visitor.visit_AsyncFunctionDef = visit_FunctionDef
+        node_visitor.visit(ast.parse(code))
+        assert expected_names == decorator_names
+    return check
+
+
+def test_get_decorator_name_simple(check_get_decorater_name):
```
jendrikseipp: decorat**o**r

tests/test_utils.py
```diff
@@ -0,0 +1,67 @@
...
+
+        def visit_FunctionDef(node):
+            for decorator in node.decorator_list:
+                decorator_names.append(utils.get_decorator_name(decorator))
+
+        node_visitor = ast.NodeVisitor()
+        node_visitor.visit_FunctionDef = visit_FunctionDef
+        node_visitor.visit_AsyncFunctionDef = visit_FunctionDef
+        node_visitor.visit(ast.parse(code))
+        assert expected_names == decorator_names
```
jendrikseipp: This test begs for testing a function with two decorators :-)

vulture/core.py
RJ722: Hmm, so are we supporting both  `@app.route` and `app.route`??

RJ722: Ahh, I got it all wrong, stupid me.

tests/test_ignore.py
```diff
@@ -0,0 +1,134 @@
+import pytest
+
+from vulture import core
+
+from . import check, skip_if_not_has_async
+
+
+@pytest.fixture
```
jendrikseipp: Can't we just use the `examine` function without the pytest.fixture? The new function can still be called `check_ignore`.

tests/test_ignore.py
```diff
@@ -0,0 +1,134 @@
...
+def check_ignore():
+    def examine(code, expected, ignore_names=[], ignore_decorators=[]):
```
jendrikseipp: We should always use the kwarg names when passing kwargs to a function. Using kwargs implies that we may reorder the kwargs later. I.e., use `check_ignore(code, ignore_names=['funny'], ignore_decorators=['f?o*', 'ba[rz]'])`. I'd just use `def check_ignore(code, ignore_names, ignore_decorators, expected):` to make things easier and explicit.

vulture/core.py
```diff
@@ -163,6 +168,10 @@ def get_set(typ):
...
+        self.ignore_decorators = [
+            decor[1:] if decor.startswith('@') else decor
+            for decor in self.ignore_decorators]
```
jendrikseipp: `[decorator.lstrip('@') for decorator in self.decorators]`

tests/test_utils.py
```diff
@@ -0,0 +1,91 @@
+import ast
+
+import pytest
+
+from . import skip_if_not_has_async
+from vulture import utils
+
+
+@pytest.fixture
+def check_get_decorator_name():
```
jendrikseipp: Same here, no need to return the function as a fixture.

tests/test_ignore.py
```diff
@@ -0,0 +1,134 @@
...
+@app.route('something')
+@foobar
+async def async_function():
+    pass
+
+@a.b.c
+async def foo():
+    pass
+"""
+    check_ignore(code, ['foo'], [], ['@app.route', '@a.b'])
```
jendrikseipp: We also need tests for these cases: 

```
@foo
@property
def bar:
    pass

--ignore-decorators @foo
-> ignore bar


@foo
@property
def bar:
    pass

--ignore-decorators @baz
-> do not ignore bar


@foo
@property
def bar:
    pass

--ignore-decorators @property
-> ignore bar
```

RJ722:
```
@foo
@property
def bar:
    pass

--ignore-decorators @property
-> ignore bar
```
This case shouldn't ignore, right??

jendrikseipp: It should. We shouldn't handle @property any differently.

tests/test_ignore.py
```diff
@@ -0,0 +1,156 @@
+from vulture import core
+
+from . import check, skip_if_not_has_async
+
+
+def check_ignore(code, expected, ignore_names, ignore_decorators):
```
jendrikseipp: Could you please change the argument order? I find it much more intuitive to use `check_ignore(code, ignore_names, ignore_decorators, expected)` (as in my earlier comment).

vulture/core.py
```diff
@@ -520,6 +539,11 @@ def csv(exclude):
         help='Comma-separated list of paths to ignore (e.g.,'
         ' "*settings.py,docs/*.py"). {glob_help} A PATTERN without globbing'
         ' characters is treated as *PATTERN*.'.format(**locals()))
+    parser.add_argument(
+        '--ignore-decorators', metavar='PATTERNS', type=csv,
+        help='Comma-separated list of decorators. Functions and classes using'
+        ' these decorators are ignored (e.g., "@app.route").'
```
jendrikseipp: "@app.route" -> "@app.route,@require_*"

tests/test_ignore.py
```diff
@@ -0,0 +1,156 @@
...
+    code = """\
+@bar
+@property
+def foo():
+    pass
+"""
+    check_ignore(code, [], [], ['@bar'])
+    check_ignore(code, ['foo'], [], ['@baz'])
+    check_ignore(code, [], [], ['property'])
+    check_ignore(code, [], [], ['@property'])
```
jendrikseipp: We need the same test function again but with the order of @bar and @property exchanged. Then the new test will fail.

bar: @jendrikseipp say what?

jendrikseipp: Hehe, sorry, I was referencing a hypothetical Python property `@bar` :-)

tests/test_ignore.py
RJ722: Done! :-)

tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
+import ast
+
+from . import skip_if_not_has_async
+from vulture import utils
+
+
+def check(code, expected_names):
```
RJ722: I think this function should be renamed to `check_decorator_names`, do you agree @jendrikseipp?

jendrikseipp: Yes, please :-)

README.rst
```diff
@@ -74,13 +74,17 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
...
 You can use ``--ignore-names foo*,ba[rz]`` to let Vulture ignore all names
-starting with ``foo`` and the names ``bar`` and ``baz``.
-
-We recommend using whitelists instead of ``--ignore-names`` whenever
-possible, since whitelists are automatically checked for syntactic
-correctness when passed to Vulture and sometimes you can even pass
-them to your Python interpreter and let it check that all whitelisted
-code actually still exists in your project.
+starting with ``foo`` and the names ``bar`` and ``baz``. Additionally, the
+``--ignore-decorators`` option should be used to ignore functions decorated
```
jendrikseipp: should -> can

README.rst
```diff
@@ -74,13 +74,17 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
...
+with the given decorator. This is helpful for example in Flask projects,
+where you can use ``--ignore-names "@app.route"`` to ignore all functions
+with the ``app.route`` decorator.
```
jendrikseipp: ``@app.route``

README.rst
```diff
@@ -74,13 +74,17 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
...
-
-We recommend using whitelists instead of ``--ignore-names`` whenever
-possible, since whitelists are automatically checked for syntactic
-correctness when passed to Vulture and sometimes you can even pass
-them to your Python interpreter and let it check that all whitelisted
-code actually still exists in your project.
+starting with ``foo`` and the names ``bar`` and ``baz``. Additionally, the
+``--ignore-decorators`` option should be used to ignore functions decorated
+with the given decorator. This is helpful for example in Flask projects,
+where you can use ``--ignore-names "@app.route"`` to ignore all functions
```
jendrikseipp: ignore-names -> ignore-decorators

README.rst
```diff
@@ -74,13 +74,17 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
...
+with the ``app.route`` decorator.
+
+We recommend using whitelists instead of ``--ignore-names`` or
+``--ignore-decorator`` whenever possible, since whitelists are automatically
```
jendrikseipp: --ignore-decorator**s**

README.rst
```diff
@@ -74,13 +74,17 @@ ignore a whole file or directory, use the ``--exclude`` parameter (e.g.,
...
+checked for syntactic correctness when passed to Vulture and sometimes
```
jendrikseipp: sometimes -> often

tests/test_ignore.py
```diff
@@ -0,0 +1,169 @@
+from vulture import core
+
+from . import check, skip_if_not_has_async
+
+
+def check_ignore(code, ignore_names, ignore_decorators, expected):
+    v = core.Vulture(
+            verbose=True, ignore_names=ignore_names,
```
jendrikseipp: Use 4 spaces to indent.

tests/test_ignore.py
```diff
@@ -0,0 +1,169 @@
...
+    pass
+"""
+    check_ignore(code, ['foo*'], [], ['bar'])
+
+
+def test_class():
+    code = """\
+class Foo:
+    def __init__(self):
+        pass
```
jendrikseipp: Let's add a second class Bar here that is not ignored. It's always good to have tests for more than once case. You can just use 
```
class Foo:
    pass

class Bar:
    pass
```

vulture/core.py
```diff
@@ -163,6 +168,9 @@ def get_set(typ):
         self.used_names = get_set('name')
 
         self.ignore_names = ignore_names or []
+        self.ignore_decorators = ignore_decorators or []
+        self.ignore_decorators = [
+            decorator.lstrip('@') for decorator in self.ignore_decorators]
```
jendrikseipp: With our current design ``--ignore-decorators "*@app.route"`` won't ignore the decorated functions even though it should. I doubt that anybody will ever use that, but it shows that maybe our design should be simplified. Let's change it so that get_decorator_name prepends an @ to the returned names and let's not strip leading @'s. Then "*@app.route", "@*app.route" and "*app.route" all work.

RJ722: Hmm, nice catch - Let me do the required. :-)

Thanks!

RJ722: > Then "*@app.route", "@*app.route" and "*app.route" all work.

But, dong this also breaks `app.route` since `fnmatchcase('app.route', '@app.route') != True`. Maybe rather than stripping `@`, we can add one if missing. What say?

RJ722: I ended up doing `'@' + decorator.lstrip('@') for decorator in ignore_decorators`. Please let me know if you see something wrong with this approach. :-)

tests/test_ignore.py
RJ722: Hey, my editor is already showing 4 space indentation.

Do you mean that we should be using the following format instead? 😅 
```python
def check_ignore(code, ignore_names, ignore_decorators, expected):
    v = core.Vulture(verbose=True, ignore_names=ignore_names,
                     ignore_decorators=ignore_decorators)
    v.scan(code)
    check(v.get_unused_code(), expected)
```

vulture/core.py
RJ722: Oops, if the initial character of the input is a wildcard (e.g., *, ?) - this would fail. :-(

tests/test_ignore.py
RJ722: Hmm, are you sure about this, because starting on column 13 on line9 makes flake8 angry:
```
tests/test_ignore.py:8:9: E128 continuation line under-indented for visual indent
```

vulture/core.py
RJ722: ahh, yeah let's do that - I like the `@` syntax anyways - more clear, less chances of error.

tests/test_ignore.py
jendrikseipp: OK, let's leave it as it is then :-)

jendrikseipp:
Thanks, Rahul!

Issue #143, opened 2018-07-27
make_whitelist is not available as an API


RJ722:
Currently, the `make_whitelist` function creates a whitelist and then prints it to stdout, whereas this function might be useful for other linters using Vulture (e.g., https://github.com/coala/coala-bears/issues/2640) and thus should be a part of the public API.

IMHO, we should return a list from the `make_whitelist` method (much like `get_unused_code`) and utilize it in the `report` method when the ``--make_whitelist`` flag is passed.

jendrikseipp:
Could you please paste an example snippet showing how you'd like to use the new API?

RJ722:
I think that `whitelist_message` can be a method of the `Item` class - This would automatically expose it as an API and then would also enable us to use it internally with the report function.

The API might then be used as following:
```python
v = vulture.Vulture()
v.scavenge(filenames)
for item in v.get_unused_code():
     print(item.whitelist_message)
```

jendrikseipp:
Sounds good. Let's call the method `Item.get_whitelist_string()` though.

PR #144, opened 2018-07-31
Use Item.get_whitelist_string.

RJ722:
## Description

Use `Item.get_whitelist_string` to create a message string.  This was previously implemented using `make_whitelist`, but that was redundant and also prevented whitelist functionality to be a part of the Vulture API.

## Related Issue

Closes: #143 

## Checklist:

### DIFF
--- a/tests/test_make_whitelist.py
+++ b/tests/test_make_whitelist.py
@@ -9,9 +9,8 @@ def check_whitelist(v):
     def examine(code, results_before, results_after):
         v.scan(code)
         check(v.get_unused_code(), results_before)
-        whitelist = [item.get_whitelist_string() for item in v.get_unused_code()]
-        for whitelist_string in whitelist:
-            v.scan(whitelist_string)
+        for item in v.get_unused_code():
+            v.scan(item.get_whitelist_string())
         check(v.get_unused_code(), results_after)
     return examine
 
--- a/tests/test_make_whitelist.py
+++ b/tests/test_make_whitelist.py
@@ -5,14 +5,13 @@ assert v  # silence pyflakes
 
 
 @pytest.fixture
-def check_whitelist(v, capsys):
+def check_whitelist(v):
     def examine(code, results_before, results_after):
         v.scan(code)
         check(v.get_unused_code(), results_before)
-        capsys.readouterr()  # Clear captured text
-        v.make_whitelist()
-        whitelist = capsys.readouterr().out
-        v.scan(whitelist)
+        whitelist = [item.get_whitelist_string() for item in v.get_unused_code()]
+        for whitelist_string in whitelist:
+            v.scan(whitelist_string)
         check(v.get_unused_code(), results_after)
     return examine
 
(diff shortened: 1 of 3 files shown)


vulture/core.py
```diff
@@ -262,24 +270,16 @@ def by_size(item):
...
+               make_whitelist=False):
         """
         Print ordered list of Item objects to stdout.
         """
         for item in self.get_unused_code(
                 min_confidence=min_confidence, sort_by_size=sort_by_size):
-            print(item.get_report(add_size=sort_by_size))
+            report = (item.get_whitelist_string() if make_whitelist
+                      else item.get_report(add_size=sort_by_size))
+            print(report)
```
jendrikseipp: No need for the `report` variable, let's just print directly.

vulture/core.py
```diff
@@ -130,6 +130,14 @@ def get_report(self, add_size=False):
             utils.format_path(self.filename), self.first_lineno,
             self.message, self.confidence, size_report)
 
+    def get_whitelist_string(self):
+        if self.typ != 'unreachable_code':
```
jendrikseipp: It's usually simpler to understand if we use a positive test in the if-statement, i.e., `if self.typ == 'unreachable_code'`.

vulture/core.py
```diff
@@ -130,6 +130,14 @@ def get_report(self, add_size=False):
...
+            prefix = '_.' if self.typ in ['attribute', 'property'] else ''
+            return "{}{}  # unused {} ({}:{:d})".format(
+                    prefix, self.name, self.typ,
+                    utils.format_path(self.filename), self.first_lineno)
+        return ''
```
jendrikseipp: Let's return a comment for unreachable code: `return '# {message} in file {filename} at line {first_lineno}'.format(**self.__dict__)`

tests/test_make_whitelist.py
```diff
@@ -5,14 +5,13 @@
...
     def examine(code, results_before, results_after):
         v.scan(code)
         check(v.get_unused_code(), results_before)
-        capsys.readouterr()  # Clear captured text
-        v.make_whitelist()
-        whitelist = capsys.readouterr().out
-        v.scan(whitelist)
+        whitelist = [item.get_whitelist_string() for item in v.get_unused_code()]
+        for whitelist_string in whitelist:
+            v.scan(whitelist_string)
```
jendrikseipp: One for-loop should be enough, no need for the generator list.

RJ722: What was I even thinking? 😅

tests/test_report.py
```diff
@@ -54,6 +53,7 @@ def test_make_whitelist(check_report):
 bar  # unused function ({filename}:7)
 _.foobar  # unused attribute ({filename}:8)
 foobar  # unused variable ({filename}:9)
+# unreachable code after 'return' in file {filename} at line 11
```
jendrikseipp: I think we should use the same format for unreachable code as in the other cases after all: `# unreachable code after 'return' ({filename}:11)`

jendrikseipp:
Thanks!

PR #145, opened 2018-08-12
.travis.yml: Test on py37, py38-dev

RJ722:
## Description

Increase support for Python3.7 and also test for py38 (allowed failures). Also, switch to `--only-binary` instead of using `--use-wheel` which has been deprecated since pip7 and removed in pip10 beta-1

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,26 @@
 language: python
-python:
-- pypy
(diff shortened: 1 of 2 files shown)


jendrikseipp:
Nice, thanks!

Issue #152, opened 2018-10-26
Ignore type checking imports


Dobatymo:
Code like this is common when type annotations in comments are used in Python:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
	from typing import Optional, Iterator
	from abc import xyz
```
These will usually show up as `unused import` as they are only referenced from comments.

The code should probably be parsed with a `TYPE_CHECKING=False` in mind.

jendrikseipp:
Could you please link to a project that uses this type of code, ideally directly to the source file?

Dobatymo:
Random large python project: https://github.com/RasaHQ/rasa_nlu/blob/master/rasa_nlu/components.py#L23-L25

jendrikseipp:
Thanks! I think it would be good to add a whitelist module for the classes (and maybe other types) in the `typing` module (https://docs.python.org/3/library/typing.html). Pull requests welcome :-) Detecting custom types (like the ones you linked above) is probably out of the scope of Vulture.

jendrikseipp:
We discussed this further in #153 and found that adding a whitelist for the typing module won't help after all:

The code linked above uses type comments. To identify that the classes are actually used in the type comments we would need to parse the comments, which I think is tedious. Here, a manual whitelist is preferable, I think. More modern type-annotated code needs Python 3.6. There you can annotate function parameters and variables. This is actually already supported by Vulture.

dsuch:
Hello,

I realise that this ticket is closed but perhaps for a future discussion, I just want to leave a note.

First, thanks for creating Vulture, this is a very useful tool!

As to the gist of this ticket - yes, code which does exactly what @Dobatymo describes actually exists and I can attest to that.

Using type hints directly in function or method definitions is not 100% the same as placing them in comments.

For instance:

```python
from typing import List

def func(a:List[str]) -> None:
  pass
```

And:

```python
if 0:
  from typing import List

def func(a):
  # type: (List[str]) -> None
  pass
```

Or:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
  from typing import List

def func(a):
  # type: (List[str]) -> None
  pass
```

The first of these incurs the most runtime overhead, the second one the least and the last one is in the middle.

Naturally, as far as type checking goes, all of the achieve the same but I am just saying that there is software that pays attention to such details and, personally, I prefer to use comments to have as little overhead as possible.

This is just one of dozens and dozens of micro-optimisations on top of various macro ones. 

[...]

PR #153, opened 2018-11-22
Ignore type checking imports

kx-chen:
## Description

Add type checking imports from `typing` to whitelist. 

File named as `typing_whitelist.py`, as it seems the code looks for those files, and it matches the other file names.

Added a `try` block around the importing, as sometimes `typing` is not always available. 

I got the types from https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html and https://docs.python.org/3/library/typing.html. Please advise if I've missed any or need any changes.

## Related Issue

https://github.com/jendrikseipp/vulture/issues/152

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/tests/test_scavenging.py
+++ b/tests/test_scavenging.py
@@ -1,3 +1,7 @@
+import sys
+
+import pytest
+
 from . import check, skip_if_not_has_async, v
 assert v  # Silence pyflakes.
 
@@ -541,3 +545,41 @@ a = 2
     check(v.defined_vars, ['a', 'a'])
     check(v.used_names, [])
     check(v.unused_vars, ['a', 'a'])
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6),
+                    reason="requires python3.6 or higher")
+def test_typing_module(v):
+    v.scan("""\
+from typing import Iterable, List
+
+x: List[int] = [1]
+
+def f(n: int) -> Iterable[int]:
+    i = 0
+    while i < n:
+        yield i
+        i += 1
+""")
+
+    check(v.unused_vars, ['x'])
+    check(v.unused_funcs, ['f'])
+    check(v.used_names, ['Iterable', 'n', 'i', 'int', 'List'])
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6),
+                    reason="requires python3.6 or higher")
+def test_type_hint_comments(v):
+    v.scan("""\
+import typing
+
+if typing.TYPE_CHECKING:
+    from typing import List, Text
+
+def foo(foo_li):
+    # type: (List[Text]) -> None
+    for bar in foo_li:
+        bar.xyz()
+""")
+
+    check(v.unused_imports, ['List', 'Text'])
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -5,6 +5,7 @@ News
 ----------------
 * Add ``sys.excepthook`` to ``sys`` whitelist.
 * Add whitelist for ``ctypes`` module.
+* Add test to check that ``typing`` module imports are marked as used
 
 
 1.0 (2018-10-23)
--- a/NEWS.rst
+++ b/NEWS.rst
@@ -5,7 +5,7 @@ News
 ----------------
 * Add ``sys.excepthook`` to ``sys`` whitelist.
 * Add whitelist for ``ctypes`` module.
-* Add test to check that ``typing`` module imports are marked as used
+* Check that Python 3.6 type-annotations are parsed and type comments are ignored (thanks @kx-chen).
 
 
 1.0 (2018-10-23)


vulture/whitelists/typing_whitelist.py
```diff
@@ -0,0 +1,14 @@
...
+    typing.Any
+    typing.Callable
+    typing.Dict
+    typing.Iterable
+    typing.List
+    typing.Optional
+    typing.Set
+    typing.Tuple
+    typing.Union
+except ImportError:
```
RJ722: Some of the classes from https://docs.python.org/3/library/typing.html#classes-functions-and-decorators are missing - for example:
```
typing.AnyStr
typing.ClassVar
typing.NoReturn
```

jendrikseipp:
Actually, before going further, I'd like to see some code where these classes are really reported as unused by Vulture. Can you provide a link, please?

kx-chen:
@jendrikseipp Sure thing. I tested the changes locally with this code mentioned in the original issue. https://github.com/RasaHQ/rasa_nlu/blob/master/rasa_nlu/components.py#L23-L25

kx-chen:
Thanks for the clarification. I believe the running vulture on the entire script above without the typing classes whitelisted should report them as unused.

Also, just to clarify again, I should be adding all the classes in https://docs.python.org/3/library/typing.html#classes-functions-and-decorators , correct? :)

RJ722:
> Thanks for the clarification. I believe the running vulture on the entire script above without the typing classes whitelisted should report them as unused.

Yes.

>  Also, just to clarify again, I should be adding all the classes in https://docs.python.org/3/library/typing.html#classes-functions-and-decorators , correct? :)

Yes, most of them. I am just a little confused about `typing.AnyStr`. Also, I'm not sure about type variables (`typing.TypeVar` and `typing.AnyStr`) because it is mandatory to instantiate them (I guess) and thus, they won't be reported by Vulture anyways.

kx-chen:
> Yes, most of them. I am just a little confused about typing.ClassVar.
> Also, I'm not sure about type variables (typing.TypeVar and typing.AnyStr) because it is mandatory to instantiate them (I guess) and thus, they won't be reported by Vulture anyways.

Yes, it looks like those classes should be instantiated, and therefore not whitelisted (I think). `Generic`, `NewType` and seem to be in the same boat too.

jendrikseipp:
I think Rahul is right, adding a whitelist for the `typing` module won't help here (my earlier stance on this in #152 was wrong). I looked into this a bit:

The code linked to in #152 uses type comments. To identify that the classes are actually used in the type comments we would need to parse the comments, which I think is tedious. Here, a manual whitelist is preferable, I think. More modern type-annotated code needs Python 3.6. There you can annotate function parameters and variables. I think this is actually something we should support.

I can't test this because I don't have Python 3.6, but I think Vulture should already see that Iterable and List are used in the following code 

```
from typing import Iterable, List

x: List[int] = [1]

def f(n: int) -> Iterable[int]:
    i = 0
    while i < n:
        yield i
        i += 1
```

Could one of you check this, please? If Vulture handles this, having a test for this would be nice (on Python 3.6+). If it doesn't, we should fix it.

jendrikseipp:
For reference, here is the pull request that added support for annotations to pyflakes: https://github.com/PyCQA/pyflakes/pull/313

RJ722:
> Yes, it looks like those classes should be instantiated, and therefore not whitelisted (I think). Generic, NewType and seem to be in the same boat too.

`Generic` - Yes, a user would need to inherit from this class in order to use this, so I guess we wouldn't be needing it. Also, `NewType` is a function so we won't be needing it. :-)

RJ722:
Hey @jendrikseipp 

For the code snippet above, Vulture outputs following:
```
type_vulture.py:3: unused variable 'x' (60% confidence)
type_vulture.py:5: unused function 'f' (60% confidence)
```

\o/

jendrikseipp:
Perfect, thanks! @kx-chen if you like, you could add a test that Vulture correctly handles type annotations. We don't need the typing whitelist, though.

kx-chen:
Sure thing. So no whitelist and to simply add tests. Would this be only on Python 3.6?

jendrikseipp:
Yes. There are already other tests that only run for certain Python versions. Maybe @RJ722  can have a look once you updated the pull request.

kx-chen:
Sorry for the delayed response, I had school in the afternoon. Currently the tests I've written run the same snippet of code as above, and asserts that the variables `x` and the function `f` is unused. Thoughts on this?

jendrikseipp:
The test has to assert that the List and Iterable imports are not reported.

kx-chen:
Sure, I'll work on it now. Anything else I should add? Is the current code snippet enough?

tests/test_scavenging.py
```diff
@@ -541,3 +545,23 @@ def test_multiple_definition(v):
...
+def f(n: int) -> Iterable[int]:
+    i = 0
+    while i < n:
+        yield i
+        i += 1
+""")
+
+    check(v.unused_vars, ['x'])
+    check(v.unused_funcs, ['f'])
+    check(v.used_names, ['Iterable', 'n', 'i', 'int', 'List'])
```
RJ722: I think it would be a good idea to add the following case as well (perhaps in a different test function):

```py
import typing

if typing.TYPE_CHECKING:
    from typing import List, Text

def foo(foo_li):
    # type: (List[Text]) -> None
    for bar in foo_li:
        bar.xyz()
```

In this snippet, you should assert that List, Text indeed are being rendered unused - just to make a point that Vulture does not support parsing of comments.

kx-chen: Sure thing. The test has been added

jendrikseipp:
Looks good. I'll make some cosmetic changes and then we can merge this.

tests/test_scavenging.py
```diff
@@ -541,3 +545,41 @@ def test_multiple_definition(v):
...
+
+
+@pytest.mark.skipif(sys.version_info < (3, 6),
```
RJ722: Hey, can we use [`pytest.importorskip`](https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency) instead of checking for Python version here?

kx-chen: With this approach, Python 3.5 and 2.7 have their tests fail for me. Using `pytest.importorskip('typing')`

RJ722: Okay, no problem - let's merge this! :-)

jendrikseipp:
Thanks, guys!

PR #165, opened 2019-07-15
Speedup creating Item objects

jogo:
According to vmprof (https://github.com/vmprof/vmprof-python) over
a large code base, about 20% of the time was spent in Item:__init__,
use slots to reduce the time spent in Item:__init__ to 1%

```import timeit


class Foo(object):
    def __init__(self):
        self.a = 1
        self.b = 2
        self.c = 3
        self.d = 4
        self.e = 5
        self.f = 6
        self.g = 7


class Bar(object):
    __slots__ = ('a', 'b', 'c', 'd', 'e', 'f', 'g')


print(timeit.timeit('Foo()', 'from __main__ import Foo'))
print(timeit.timeit('Bar()', 'from __main__ import Bar'))
```

```
0.796907901764
0.0809969902039
```

## Checklist:

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -102,6 +102,8 @@ class Item(object):
     """
     Hold the name, type and location of defined code.
     """
+    __slots__ = ('name', 'typ', 'filename', 'first_lineno',
+                 'last_lineno', 'message', 'confidence')
 
     def __init__(self, name, typ, filename, first_lineno, last_lineno,
                  message='',


jogo:
Not sure if the failure here https://ci.appveyor.com/project/jendrikseipp96693/vulture/builds/25999090 is related or not

jendrikseipp:
Nice PR! Do you have before and after time results for your code base?

I think the AppVeyor failed build is unrelated to your changes.

PR #168, opened 2019-08-30
Fix Windows Build

RJ722:
## Description

Previously, pip was too old and did not understand pip's metadata. Updating virtualenv before running tests solves this problem.

Also, add tox environments to Appveyor (as recommended on [their website][appv]).

[appv]: https://www.appveyor.com/docs/lang/python/

## Checklist:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,21 @@
 version: 1.0.{build}
 
+environment:
+  matrix:
+  - TOXENV: cleanup
+  - TOXENV: docs
+  - TOXENV: py27
+    PYTHON: "C:\\Python27-x64"
+  - TOXENV: py35
+    PYTHON: "C:\\Python35-x64"
+  - TOXENV: py36
+    PYTHON: "C:\\Python36-x64"
+  - TOXENV: py37
+    PYTHON: "C:\\Python37-x64"
+  - TOXENV: style
+
 install:
-- cmd: python -m pip install -U pip setuptools tox wheel
+- cmd: python -m pip install -U pip setuptools tox wheel virtualenv
 
 build_script:
 - cmd: python setup.py bdist_wheel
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -1,7 +1,21 @@
...
+  - TOXENV: py27
+    PYTHON: "C:\\Python27-x64"
+  - TOXENV: py35
+    PYTHON: "C:\\Python35-x64"
+  - TOXENV: py36
+    PYTHON: "C:\\Python36-x64"
+  - TOXENV: py37
+    PYTHON: "C:\\Python37-x64"
+  - TOXENV: style
+
```
jendrikseipp: What's the advantage of specifying the tox environments here? This has the disadvantage that we need to maintain this list of environments in an additional file and I can see no real benefit.

RJ722: I did this because it was recommended on [Appveyor's website](https://www.appveyor.com/docs/lang/python/):

> For any Tox environment you want to run on AppVeyor you need to add a `TOXENV`: ... item to your environment matrix in appveyor.yml.

After looking into it, I find that it doesn't do much other than splitting the CI into multiple jobs, and for each env, repeating all the instructions in `appveyor.yml`. But, this actually helps us build wheels for all Python versions and test on them.

Also, I see that [tox's `envlist`](https://tox.readthedocs.io/en/latest/config.html#conf-envlist) first looks up for `TOXENV` variable, and then the `tox.ini`:

> Determining the environment list that tox is to operate on happens in this order (if any is found, no further lookups are made):
> * command line option `-eENVLIST`
> * environment variable `TOXENV`
> * tox.ini file’s envlist

However, I am not sure what to make of that! 😅

jendrikseipp: Ah, thanks for investigating! I think having the TOXENVs in the appveyor file makes sense, since appveyor is the one who knows which environments there are.

PR #169, opened 2019-08-31
Update CI

jendrikseipp:


appveyor.yml
```diff
@@ -2,8 +2,6 @@ version: 1.0.{build}
 
 environment:
   matrix:
-  - TOXENV: cleanup
-  - TOXENV: docs
```
RJ722: Sure, we shouldn't be checking for `docs`?

jendrikseipp: This is what the tox -e ..,docs,... line does below.

RJ722: ahh, of course. thanks for the clarification! :-)

Issue #175, opened 2019-10-21
visit_Str is deprecated in Python3.8


RJ722:
According to https://docs.python.org/3/library/ast.html:

> Deprecated since version 3.8: Methods visit_Num(), visit_Str(), visit_Bytes(), visit_NameConstant() and visit_Ellipsis() are deprecated now and will not be called in future Python versions. Add the visit_Constant() method to handle all constant nodes.

Of all of these, we are only using the `visit_Str` function.

jendrikseipp:
Interesting! It seems ast.Constant was introduced in Python 3.6. I think we should add a visit_Constant() method alongside visit_Str() to be future-proof.

PR #176, opened 2019-10-23
Update core.py

ssurbhi560:
## Description

Since visit_Str() is deprecated in Python3.8 so adding the visit_Constant() method .  #175 

## Checklist:

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -397,7 +397,8 @@ class Vulture(ast.NodeVisitor):
                 self.defined_classes, node.name, node, ignore=_ignore_class)
     
     def visit_Constant(self, node):
-        self.visit_Str(node)
+        if getattr(node, 's'):
+            self.visit_Str(node)
 
     def visit_FunctionDef(self, node):
         decorator_names = [utils.get_decorator_name(
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -395,6 +395,9 @@ class Vulture(ast.NodeVisitor):
         else:
             self._define(
                 self.defined_classes, node.name, node, ignore=_ignore_class)
+    
+    def visit_Constant(self, node):
+        self.visit_Str(node)
 
     def visit_FunctionDef(self, node):
         decorator_names = [utils.get_decorator_name(
--- a/vulture/whitelists/ast_whitelist.py
+++ b/vulture/whitelists/ast_whitelist.py
@@ -17,6 +17,7 @@ whitelist_node_visitor.visit_Bytes
 whitelist_node_visitor.visit_Call
 whitelist_node_visitor.visit_ClassDef
 whitelist_node_visitor.visit_Compare
+whitelist_node_visitor.visit_Constant
 whitelist_node_visitor.visit_Delete
 whitelist_node_visitor.visit_Dict
 whitelist_node_visitor.visit_DictComp


vulture/core.py
```diff
@@ -395,6 +395,9 @@ def visit_ClassDef(self, node):
         else:
             self._define(
                 self.defined_classes, node.name, node, ignore=_ignore_class)
+    
+    def visit_Constant(self, node):
```
jendrikseipp: Is this method really called at all? node.s is accessed in visit_Str and I doubt that all constants have this field.

RJ722: > Is this method really called at all?

I just tried a basic example:

```python
class CustomNodeVisitor(ast.NodeVisitor):
    def scan(self, code):
        self.visit(ast.parse(code))

    def visit_Constant(self, node):
        print("I'm visiting a `Constant` node! \o/")

code = """\
import this
username = "RJ722"
"""

c = CustomNodeVisitor()
c.scan(code)
```

which has the following output:
```
I'm visiting a `Constant` node! \o/
```

so I guess it is being called.

>  I doubt that all constants have this field.

Yeah, maybe we can try something like:

```python
def visit_Constant(self, node):
    if getattr(node, 's'):
        self.visit_Str(node)
```

However, I'm not sure that this is correct!

RJ722:
Thank you @ssurbhi560 for the PR!

Your tests are failing because currently Vulture is reporting `visit_Constant` method as unused (because it's dynamically called).

You can add the following in `vulture/whitelists/ast_whitelist.py` and the tests would go green again:

```python
whitelist_node_visitor.visit_Constant
```

However, I'm not entirely sure how we can add a test case for this new functionality.

vulture/core.py
RJ722: I edited my previous comment to simplify the test case:

Initially, it was:

```python
import this
username = 'RJ722'
```

which has a redundant import statement and an unused variable. We can just use:

```
'RJ722'
```

vulture/core.py
```diff
@@ -395,6 +395,10 @@ def visit_ClassDef(self, node):
...
+        if getattr(node, 's'):
+            self.visit_Str(node)
```
RJ722: I think we should do `getattr(node, 's', None)` otherwise it will throw an error whenever it won't find `'s'` as an attribute.

Other than that, it looks good to me.

jendrikseipp: `hasattr(node, 's')` would be even better.

But the Constant node never has a `s` attribute. I think the correct way would be:

```
def visit_Constant(self, node):
    if isinstance(node.value, str):
        self._handle_string(node.value)
```

Also, the visit_Str method should only be defined for Python < 3.8. Once we do that, it will be easier to test the new code.

RJ722: > hasattr(node, 's') would be even better.

Oh, of course! 😅

> Also, the visit_Str method should only be defined for Python < 3.8. Once we do that, it will be easier to test the new code.

Hmm, interesting!

I'm not sure how we can conditionally define a function in Python, but I am thinking somewhere along the lines of:

```python
def define_if_has_Str(f):
    if sys.version_info >= (3, 8):
        return f
    del f    # Not sure if it deletes the global reference or just locally
    def _f(*args):
        pass
    return _f


class Vulture:
    # ...

    @define_if_has_Str
    def visit_Str(self, node):
        self._handle_string(node.s)
```

@jendrikseipp is this too hacky?

What all other options do we have otherwise?

jendrikseipp:
I took the liberty of fixing this here: https://github.com/jendrikseipp/vulture/commit/c5b6417a03736fc25736ea1dcd23385c7c462089

Issue #190, opened 2020-02-27
"Unreachable code after _name_" entries cannot be used in whitelist


HitLuca:
After running vulture in order to generate a whitelist with

`vulture my_module --make-whitelist > whitelist.py`

I noticed that a class of entries cannot silence the script's output (`vulture my_module whitelist.py`), specifically the unreachable code ones.

`# unreachable code after '_name_' (my_module/.../name.py:###)`

I didn't find a way to reliably suppress such outputs when running vulture, I tried to remove the comment `#`, but that breaks the whitelist parsing.

Not that it matters too much for me, I only have 3 entries of that sort, but for large projects it could be a bigger issue.

jendrikseipp:
Thanks for the report! Unfortunately, there is no way to silence such errors at the moment. The assumption is that unreachable code should normally be easy to be removed. Is that different in your code? Can you give an example of where this happens?

HitLuca:
My use case is of a coworker that is adding some functionality but at the moment it's not fully completed. For that reason he returns early or raises an exception, but after it there's code that needs to be there for a future feature. Whether is bad practice or not I believe there should be an option to stop such warnings

jendrikseipp:
We'd like to honor `# noqa` comments eventually, but so far no one has made a pull request. See also #160.

RJ722:
I think we can close this.

jendrikseipp:
Yes, that's right.

Issue #182, opened 2019-12-11
Change the whitelist file extension in readme


impredicative:
The example `vulture mydir --make-whitelist > whitelist.py` in the readme typically creates an invalid Python file since it lacks imports. Such a file would create a problem for multiple other static analyzers that would find it to be grossly invalid Python code. It would be better if this example were updated to to use the extension `.txt` instead as below:

```bash
$ vulture mydir --make-whitelist > vulture.txt
$ vulture mydir vulture.txt
```

The `# Vulture whitelist:` header can be added to the created file. In this way, the Python imports can continue to be missing in this whitelist file, and the other static analyzers won't have any issue with it.

---

I'm actually quite displeased with the current end-user whitelist setup. It doesn't seem well thought through at all. If I use a .py whitelist file with the appropriate imports, then vulture complains that the imports are unused. If I use a .txt whitelist file, then it's not possible to define which file each exclusion is supposed to have been imported from, making it a non-specific exclusion. If for example, I specify "my_foo_pkg.my_sub_pkg.xyzz" in a .txt file, then "my_foo_pkg.my_sub_pkg" are not checked.

RJ722:
Hello @impredicative,

Thanks for the suggestions.

> If I use a .txt whitelist file, then it's not possible to define which file each exclusion is supposed to have been imported from, making it a non-specific exclusion. If for example, I specify "my_foo_pkg.my_sub_pkg.xyzz" in a .txt file, then "my_foo_pkg.my_sub_pkg" are not checked.

This is actually much more problematic, even with `.py` extension, and part of it has to do with the fact that as Vulture parses the `ast`, it only uses the "names" of the items for comparison, without any information about the ancestry of that node -- https://github.com/jendrikseipp/vulture/blob/master/vulture/core.py#L50-L54

Let me give a more concrete example of this problem:

```
├── foo
│   ├── __init__.py
│   ├── bar
│   │   ├── __init__.py
│   │   └── ex1.py
│   └── ex2.py
└── whitelist.py
```

```python
# foo/bar/ex1.py
def unused_func():
    pass
```

```python
# foo/ex2.py
def unused_func();
    pass
```

Running `vulture foo` gives the following results:

```
foo/bar/ex1.py:1: unused function 'unused_func' (60% confidence)
foo/ex2.py:1: unused function 'unused_func' (60% confidence)
```

For ignoring only `foo.bar.ex1.unused_func`, whitelist file would look like:

```python
# whitelist.py
[...]

Issue #160, opened 2019-05-27
Honour "# noqa: F401" comments


OddBloke:
Our codebase has to work as far back as Python 3.4, which doesn't have the typing module available. As such, all of our type annotations are in strings (so they are lazily evaluated during type checking and ignored when running normally) and our typing imports are conditional, with "# noqa: F401" appended to silence the flake8 warnings.

vulture doesn't honour these comments, and it seems like an obvious win for it to do so.

(Per http://flake8.pycqa.org/en/2.5.5/warnings.html, F401 is specifically the warning for "`module` imported but unused"; this matches very closely with Vulture's unused import warnings.)

OddBloke:
Looking through existing issues, I've seen that "# noqa" support has generally been rejected. Hopefully this request for just one specific form of noqa comment will be considered, as in my opinion, it would reduce the barrier to adoption substantially.

jendrikseipp:
Since Python 3.4 has reached EOL and Vulture stopped testing for this version recently, I don't think Vulture should add workarounds that are only needed for this version. Your problem seems to be more general though. However, I don't understand why you need imports if all your type annotations are in strings. Could you please provide some example code?

anowlcalledjosh:
Even if type hints are specified as strings, you still have to import the things used in type hints to tell the type checker what they are – it doesn't matter if you say `foo: MyObject` or `foo: "MyObject"` or `foo = None  # type: MyObject`, if mypy doesn't know which `MyObject` you're referring to.

jendrikseipp:
Thanks for the clarification! I agree that honoring F401 noqa comments makes sense. I'd be happy to review a pull request that makes this change.

RJ722:
Regarding honoring `# noqa` comments, I think that once we have Vulture available as a flake-8 plugin (#161), we wouldn't need to develop/maintain this functionality ourselves. Am I thinking in the right direction?

The-Compiler:
@RJ722 I think you're right!

jendrikseipp:
That's true. The question is whether we want to offer this functionality 
to users of standalone Vulture as well. I have no strong opinion on 
this, but I think the decision should depend on the complexity of the 
patch for this functionality and how many users would like to have it.

RJ722:
@OddBloke are you okay with changing all of `# noqa: F401, V104`? (`V104` is issue code Vulture raises for unused imports)

jendrikseipp:
[...]

PR #195, opened 2020-03-09
Ignore results annotated with '# noqa'

RJ722:
Hey, 

## Description

This PR isn't production-ready by any means, but I just wanted to have an idea about the complexity it would add to Vulture to support `# noqa` comments. I'd be happy to add tests and documentation, once we all settle if it would be a good idea to carry this forward.

Another change with the introduction of `# noqa` comments would be the introduction of issue codes (it's possible without these as well, but would hurt specificity). For now, I've just added `V001`, `V002`, ... - the nomenclature for these issue codes is another thing to think about.

## Related Issue

#160, #182, #190 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Unreleased
+
+* Report issue codes in ouput (e.g., `code.py:1: V104 unused import ...`)
+  (RJ722, #195)
+* Support `# noqa` comments to suppress results on that line. (RJ722, #195).
+
 # 1.4 (2020-03-30)
 
 * Ignore unused import statements in `__init__.py` (RJ722, #192).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-* Report issue codes in ouput (e.g., `code.py:1: V104 unused import ...`)
+* Report issue codes in output (e.g., `code.py:1: V104 unused import ...`)
   (RJ722, #195)
 * Support `# noqa` comments to suppress results on that line. (RJ722, #195).
 
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -1,6 +1,6 @@
 import pytest
 
-from vulture.core import NOQA_REGEXP, _parse_error_codes
+from vulture.noqa import NOQA_REGEXP, parse_error_codes
 from . import check, v
 
 assert v  # Silence pyflakes.
@@ -19,7 +19,7 @@ assert v  # Silence pyflakes.
 )
 def test_noqa_regex_present(line, codes):
     match = NOQA_REGEXP.search(line)
-    parsed = _parse_error_codes(match)
+    parsed = parse_error_codes(match)
     assert parsed == codes
 
 
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -150,30 +150,55 @@ import dis  # noqa: V001 (code for unused attr)
     check(v.unused_imports, ["foo", "zoo", "dis"])
 
 
-def test_noqa_propoerties(v):
+def test_noqa_properties(v):
     v.scan(
         """\
 class Zoo:
-    @property  # noqa
-    def no_of_coalas(self):
-        pass
-
-    @property  # noqa: V005
-    def area(self, width, depth):  
+    @property
+    def no_of_coalas(self):  # noqa
         pass
 
     @property
-    def entry_gates(self):  # noqa
+    def area(self, width, depth):  # noqa: V005
         pass
 
-    @property  # noqa: V003 (code for unused function)
-    def tickets(self):
+    @property  # noqa  (should not be ignored)
+    def entry_gates(self):
+        pass
+
+    @property  
+    def tickets(self):  # noqa: V003 (code for unused function)
         pass
 """
     )
     check(v.unused_props, ["entry_gates", "tickets"])
     check(v.unused_classes, ["Zoo"])
     check(v.unused_vars, ["width", "depth"])
+
+
+def test_noqa_properties_multiple_decorators(v):
+    v.scan(
+        """\
+class Foo:
+    @property
+    @make_it_cool
+    @log
+    def something(self):  # noqa: V005
+        pass
+
+    @coolify
+    @property
+    def something_else(self):  # noqa: V005
+        pass
+
+    @a
+    @property
+    @b  # noqa
+    def abcd(self):
+        pass
+"""
+    )
+    check(v.unused_props, ['abcd'])
 
 
 def test_noqa_unreacahble_code(v):
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -30,15 +30,15 @@ def test_noqa_regex_present(line, codes):
     [("# noqa: 123V"), ("# noqa explaination: V012"), ("# noqa: ,V001"),],
 )
 def test_noqa_regex_no_groups(line):
-    assert NOQA_REGEXP.search(line).groupdict()["codes"] == None
+    assert NOQA_REGEXP.search(line).groupdict()["codes"] is None
 
 
 @pytest.mark.parametrize(
     "line",
-    [("#noqa"), ("##noqa"), ("# n o q a"), ("#NOQA"), ("# Hello, noqa"),],
+    [("#noqa"), ("##noqa"), ("# n o q a"), ("#NOQA"), ("# Hello, noqa")],
 )
 def test_noqa_regex_not_present(line):
-    assert bool(NOQA_REGEXP.search(line)) == False
+    assert bool(NOQA_REGEXP.search(line)) is False
 
 
 def test_noqa_attributes(v):
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -25,7 +25,13 @@ def test_noqa_regex_present(line, codes):
 
 @pytest.mark.parametrize(
     "line",
-    [("# noqa: 123V"), ("# noqa explaination: V012"), ("# noqa: ,V101")],
+    [
+        ("# noqa: 123V"),
+        ("# noqa explaination: V012"),
+        ("# noqa: ,V101"),
+        ("# noqa: #noqa: V102"),
+        ("# noqa: # noqa: V102"),
+    ],
 )
 def test_noqa_regex_no_groups(line):
     assert NOQA_REGEXP.search(line).groupdict()["codes"] is None
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -243,13 +243,3 @@ import this  # V098, A123, F876
 """
     )
     check(v.unused_imports, ["this"])
-
-
-def test_noqa_with_special_unicode(v):
-    v.scan(
-        """\
-import abc  # noqa: V012, V😎12
-import problems  # noqa: V03🙃1
-"""
-    )
-    check(v.unused_imports, ["abc", "problems"])
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -27,7 +27,7 @@ def test_noqa_regex_present(line, codes):
     "line",
     [
         ("# noqa: 123V"),
-        ("# noqa explaination: V012"),
+        ("# noqa explanation: V012"),
         ("# noqa: ,V101"),
         ("# noqa: #noqa: V102"),
         ("# noqa: # noqa: V102"),
@@ -42,7 +42,7 @@ def test_noqa_regex_no_groups(line):
     [("#noqa"), ("##noqa"), ("# n o q a"), ("#NOQA"), ("# Hello, noqa")],
 )
 def test_noqa_regex_not_present(line):
-    assert bool(NOQA_REGEXP.search(line)) is False
+    assert not NOQA_REGEXP.search(line)
 
 
 def test_noqa_without_codes(v):
@@ -52,8 +52,8 @@ import this  # noqa
 
 class Cellar:  # noqa
     @property
-    def wine(self, grapes): # noqa
-        pass
+    def wine(self): # noqa
+        grapes = True  # noqa
 
     def serve(self, quantity=50):  # noqa
         self.quantity_served = quantity  # noqa
@@ -77,8 +77,8 @@ import this  # noqa: V104
 
 class Cellar:  # noqa: V102
     @property
-    def wine(self, grapes):  # noqa: V105, V106
-        pass
+    def wine(self):  # noqa: V105
+        grapes = True  # noqa: V106
 
     def serve(self, quantity=50):  # noqa: V103
         self.quantity_served = quantity  # noqa: V101
@@ -131,7 +131,6 @@ def play(tune, instrument='bongs', _hz='50'):  # noqa: V103
 
 # noqa
 def problems():  # noqa: V104
-    ''' They don't go away. :-)'''
     pass  # noqa: V103
 
 def hello(name):  # noqa
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -1,6 +1,44 @@
+import pytest
+
+from vulture.core import NOQA_REGEXP
 from . import check, v
 
 assert v  # Silence pyflakes.
+
+
+@pytest.mark.parametrize(
+    "line, codes",
+    [
+        ("# noqa", ["all"]),
+        ("## noqa", ["all"]),
+        ("# noqa Hi, go on.", ["all"]),
+        ("# noqa: V001", ["V001"]),
+        ("# noqa: V001, V007", ["V001", "V007"]),
+        ("# NoQA: V001,      V003, \t V004", ["V001", "V003", "V004"]),
+    ],
+)
+def test_noqa_regex_present(line, codes):
+    match = NOQA_REGEXP.search(line)
+    parsed = [
+        c.strip() for c in (match.groupdict()["codes"] or "all").split(",")
+    ]
+    assert parsed == codes
+
+
+@pytest.mark.parametrize(
+    "line",
+    [("# noqa: 123V"), ("# noqa explaination: V012"), ("# noqa: ,V001"),],
+)
+def test_noqa_regex_no_groups(line):
+    assert NOQA_REGEXP.search(line).groupdict()["codes"] == None
+
+
+@pytest.mark.parametrize(
+    "line",
+    [("#noqa"), ("##noqa"), ("# n o q a"), ("#NOQA"), ("# Hello, noqa"),],
+)
+def test_noqa_regex_not_present(line):
+    assert bool(NOQA_REGEXP.search(line)) == False
 
 
 def test_noqa_attributes(v):
@@ -85,8 +123,8 @@ class Zoo:
         pass
 """
     )
-    check(v.unused_props, ['entry_gates', 'tickets'])
-    check(v.unused_classes, ['Zoo'])
+    check(v.unused_props, ["entry_gates", "tickets"])
+    check(v.unused_classes, ["Zoo"])
     check(v.unused_vars, ["width", "depth"])
 
 
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -50,12 +50,14 @@ def test_noqa_without_codes(v):
         """\
 import this  # noqa
 
-class Cellar:  # noqa
-    @property
-    def wine(self): # noqa
+@underground  # noqa
+class Cellar:
+    @property  # noqa
+    def wine(self):
         grapes = True  # noqa
 
-    def serve(self, quantity=50):  # noqa
+    @without_ice  # noqa
+    def serve(self, quantity=50):
         self.quantity_served = quantity  # noqa
         return
         self.pour()  # noqa
@@ -75,12 +77,14 @@ def test_noqa_specific_issue_codes(v):
         """\
 import this  # noqa: V104
 
-class Cellar:  # noqa: V102
-    @property
-    def wine(self):  # noqa: V105
+@underground  # noqa: V102
+class Cellar:
+    @property  # noqa: V105
+    def wine(self):
         grapes = True  # noqa: V106
 
-    def serve(self, quantity=50):  # noqa: V103
+    @without_ice  # noqa: V103
+    def serve(self, quantity=50):
         self.quantity_served = quantity  # noqa: V101
         return
         self.pour()  # noqa: V201
@@ -160,35 +164,36 @@ def test_noqa_properties(v):
         """\
 class Zoo:
     @property
-    def no_of_coalas(self):  # noqa
+    def no_of_koalas(self):  # noqa
         pass
 
     @property
     def area(self, width, depth):  # noqa: V105
         pass
 
-    @property  # noqa  (should not be ignored)
+    @property  # noqa
     def entry_gates(self):
         pass
 
-    @property
-    def tickets(self):  # noqa: V103 (code for unused function)
+    @property  # noqa: V103 (code for unused function)
+    def tickets(self):
         pass
 """
     )
-    check(v.unused_props, ["entry_gates", "tickets"])
+    check(v.unused_props, ["no_of_koalas", "area", "tickets"])
     check(v.unused_classes, ["Zoo"])
     check(v.unused_vars, ["width", "depth"])
 
 
-def test_noqa_properties_multiple_decorators(v):
+def test_noqa_multiple_decorators(v):
     v.scan(
         """\
+@bar  # noqa: V102
 class Foo:
-    @property
+    @property  # noqa: V105
     @make_it_cool
     @log
-    def something(self):  # noqa: V105
+    def something(self):
         pass
 
     @coolify
@@ -203,7 +208,8 @@ class Foo:
         pass
 """
     )
-    check(v.unused_props, ["abcd"])
+    check(v.unused_props, ["something_else", "abcd"])
+    check(v.unused_classes, [])
 
 
 def test_noqa_unreacahble_code(v):
--- a/tests/test_noqa.py
+++ b/tests/test_noqa.py
@@ -1,6 +1,6 @@
 import pytest
 
-from vulture.noqa import NOQA_REGEXP, parse_error_codes
+from vulture.noqa import NOQA_REGEXP, _parse_error_codes
 from . import check, v
 
 assert v  # Silence pyflakes.
@@ -19,7 +19,7 @@ assert v  # Silence pyflakes.
 )
 def test_noqa_regex_present(line, codes):
     match = NOQA_REGEXP.search(line)
-    parsed = parse_error_codes(match)
+    parsed = _parse_error_codes(match)
     assert parsed == codes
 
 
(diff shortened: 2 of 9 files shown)


vulture/core.py
```diff
@@ -46,6 +47,39 @@
...
+    # `# noqa:E123,W451,F921`
+    # `# NoQA: E123,W451,F921`
+    # `# NOQA: E123,W451,F921`
+    # `# NOQA:E123,W451,F921`
+    # We do not want to capture the ``: `` that follows ``noqa``
+    # We do not care about the casing of ``noqa``
+    # We want a comma-separated list of errors
+    # https://regex101.com/r/4XUuax/3 full explenation of the regex
+    r"#( )?noqa(?::[\s]?(?P<codes>([A-Z]+[0-9]+(?:[,\s]+)?)+))?",
+    re.IGNORECASE,
```
RJ722: This expression was taken from https://gitlab.com/pycqa/flake8//blob/0ac337608b3d88079aa91bdcffed7b4df6fd28a8/src/flake8/defaults.py#L26 with an added case of `#noqa` (without a space b/w `#` and `noqa`) -- explanation URL updated.

jendrikseipp: I think it makes sense to follow exactly what flake8 uses.

vulture/core.py
```diff
@@ -129,6 +164,7 @@ def __init__(
         self.last_lineno = last_lineno
         self.message = message or "unused {typ} '{name}'".format(**locals())
         self.confidence = confidence
+        self.issue_code = ISSUE_CODE_MAP[typ]
```
jendrikseipp: No need to store this here if it's easily deducible from the `typ`.

vulture/core.py
```diff
@@ -201,14 +237,27 @@ def get_set(typ):
...
         self.ignore_names = ignore_names or []
         self.ignore_decorators = ignore_decorators or []
+        self.noqa_matches = defaultdict(list)
 
         self.filename = ""
         self.code = []
         self.found_dead_code_or_error = False
 
+    def add_noqa_match(self, code, lineno):
+        self.noqa_matches[code].append(lineno)
```
jendrikseipp: No separate method needed.

vulture/core.py
```diff
@@ -46,6 +47,39 @@
...
+    # We do not want to capture the ``: `` that follows ``noqa``
+    # We do not care about the casing of ``noqa``
+    # We want a comma-separated list of errors
+    # https://regex101.com/r/4XUuax/3 full explenation of the regex
+    r"#( )?noqa(?::[\s]?(?P<codes>([A-Z]+[0-9]+(?:[,\s]+)?)+))?",
+    re.IGNORECASE,
+)
+
+
+def _find_noqa(line):
```
jendrikseipp: No separate method needed.

vulture/core.py
```diff
@@ -201,14 +237,27 @@ def get_set(typ):
...
 
         self.filename = ""
         self.code = []
         self.found_dead_code_or_error = False
 
+    def add_noqa_match(self, code, lineno):
+        self.noqa_matches[code].append(lineno)
+
+    def parse_noqa(self, code):
+        for index, line in enumerate(code):
```
jendrikseipp: enumerate(code, start=1)

vulture/core.py
```diff
@@ -201,14 +237,27 @@ def get_set(typ):
...
+            match = _find_noqa(line)
+            if match is not None:
```
jendrikseipp: `if match`

vulture/core.py
```diff
@@ -201,14 +237,27 @@ def get_set(typ):
...
+                code = match.groupdict()["codes"] or "all"
```
jendrikseipp: This needs to handle the case when multiple codes are given. (split by commas and strip whitespace, I guess.)

vulture/core.py
```diff
@@ -282,6 +331,16 @@ def get_unused_code(self, min_confidence=0, sort_by_size=False):
...
         Return ordered list of unused Item objects.
         """
+
+        def _has_noqa(obj):
+            code = (
+                obj.issue_code
+                if obj.issue_code in self.noqa_matches
+                else "all"
+            )
+            if code in self.noqa_matches:
```
jendrikseipp: I think this line (the if-test) can be removed.

vulture/core.py
```diff
@@ -201,14 +237,27 @@ def get_set(typ):
 
         self.ignore_names = ignore_names or []
         self.ignore_decorators = ignore_decorators or []
+        self.noqa_matches = defaultdict(list)
```
jendrikseipp: This should be a set instead of a list.

vulture/core.py
```diff
@@ -301,8 +360,14 @@ def by_size(item):
             + self.unreachable_code
         )
 
+        results_without_noqa = [
+            obj for obj in unused_code if not _has_noqa(obj)
+        ]
```
jendrikseipp: No need to create a filtered list. Simply filter objects below.

vulture/core.py
```diff
@@ -282,6 +331,16 @@ def get_unused_code(self, min_confidence=0, sort_by_size=False):
...
         """
+
+        def _has_noqa(obj):
+            code = (
+                obj.issue_code
+                if obj.issue_code in self.noqa_matches
+                else "all"
+            )
+            if code in self.noqa_matches:
+                return obj.first_lineno in self.noqa_matches[code]
```
jendrikseipp: I think the whole function can be reduced to `return obj.first_lineno in self.noqa_matches["all"] or obj.first_lineno in self.noqa_matches[ISSUE_CODE_MAP[obj.typ]]`

vulture/core.py
RJ722: Okay. Changed it back.

RJ722:
@jendrikseipp should `ISSUE_CODE_MAP` be a class variable for `Vulture`?

I'm asking because my guess is that most folks would only import Vulture, if using the API (`from vulture import Vulture`).

vulture/core.py
RJ722: I ended up with:

```py
return obj.first_lineno in (
    self.noqa_matches[ISSUE_CODE_MAP[obj.typ]]
    or self.noqa_matches["all"]
)
```

vulture/core.py
RJ722: Yeah, that improves readability. thanks!

vulture/core.py
RJ722: Done!

vulture/core.py
RJ722: thanks!

vulture/core.py
```diff
@@ -46,6 +47,34 @@
 if sys.version_info < (3, 4):
     IGNORED_VARIABLE_NAMES |= {"True", "False"}
 
+ISSUE_CODE_MAP = {
```
jendrikseipp: Let's call this dict ERROR_CODES. I think leaving this as a global variable is fine.

vulture/core.py
```diff
@@ -282,6 +325,13 @@ def get_unused_code(self, min_confidence=0, sort_by_size=False):
         """
         Return ordered list of unused Item objects.
         """
+
+        def _has_noqa(obj):
+            return obj.first_lineno in (
+                self.noqa_matches[ISSUE_CODE_MAP[obj.typ]]
+                or self.noqa_matches["all"]
+            )
```
jendrikseipp: This will fail if, e.g., "noqa V001" is given for line 10, "noqa" is given for line 5 and the object appears on line 5.

vulture/core.py
```diff
@@ -282,6 +325,13 @@ def get_unused_code(self, min_confidence=0, sort_by_size=False):
         """
         Return ordered list of unused Item objects.
         """
+
+        def _has_noqa(obj):
```
jendrikseipp: No reason for leading underscore in nested function.

vulture/core.py
```diff
@@ -46,6 +47,34 @@
...
+    # `# noqa: E123`
+    # `# noqa: E123,W451,F921`
+    # `# noqa:E123,W451,F921`
+    # `# NoQA: E123,W451,F921`
+    # `# NOQA: E123,W451,F921`
+    # `# NOQA:E123,W451,F921`
+    # We do not want to capture the ``: `` that follows ``noqa``
+    # We do not care about the casing of ``noqa``
+    # We want a comma-separated list of errors
+    # https://regex101.com/r/4XUuax/2 full explanation of the regex
```
jendrikseipp: No need to copy the whole comment. The first three lines and a well-formed example are enough.

vulture/core.py
```diff
@@ -201,14 +230,28 @@ def get_set(typ):
...
         self.found_dead_code_or_error = False
 
+    def parse_noqa(self, code):
+        for lineno, line in enumerate(code, start=1):
+            match = NOQA_REGEXP.search(line)
+            if match:
+                codes = [
+                    c.strip()
+                    for c in (match.groupdict()["codes"] or "all").split(",")
+                    # if no code is specified, assign it to `all`
```
jendrikseipp: Comment above code.

vulture/core.py
```diff
@@ -301,8 +351,11 @@ def by_size(item):
             + self.unreachable_code
         )
 
+        # breakpoint()
         confidently_unused = [
```
jendrikseipp: Rename to "reported_unused".

vulture/core.py
RJ722: Oh yes, nice catch. We need a union of two sets here.

RJ722:
Another thing I was pondering about. We don't start ignoring `noqa` until `get_unused_code` is called. What if the user bypasses it, e.g.:

```py
from vulture import Vulture

v = Vulture()
v.scan('import this  # noqa: V004')
print(v.unused_imports)
```

Should we move the `has_noqa` logic to `_define` method?

jendrikseipp:
> Should we move the `has_noqa` logic to `_define` method?

Good idea! This aligns better with the existing code for ignoring things.

RJ722:
TODO:

- [ ] One test is failing because Python3.8 onwards, the first line for property node is the same line at which the function starts.
- [ ] Write documentation for these changes -- @jendrikseipp can you take this on, please? 😅

tests/test_noqa.py
```diff
@@ -0,0 +1,205 @@
...
+"""
+    )
+    check(v.unused_imports, ["this"])
+
+
+def test_noqa_with_special_unicode(v):
+    v.scan(
+        """\
+import abc  # noqa: V012, V😎12
+import problems  # noqa: V03🙃1
```
RJ722: This test is the culprit behind tests failing for Python2.7. Should I remove them?

jendrikseipp: Yes, I think it's easiest to remove the emojis.

jendrikseipp:
Could you also add a test that verifies that we fixed the "or vs. union" bug?

RJ722:
> Could you also add a test that verifies that we fixed the "or vs. union" bug?

The way I've added the tests is basically ensuring that an example of the same issue code is  covered in both categories - `all` and `V00X` (on different lines - both before and after). If there'd have been a problem with "or v/s union", we'd have seen it by now.

RJ722:
But, now that I come think of it, I haven't added isolated cases - just `V00X`, or `# noqa` alone. Adding them.

vulture/core.py
```diff
@@ -457,26 +495,41 @@ def _define(
...
+        def ignored(name):
+            return (ignore and ignore(self.filename, name)) or _match(
+                name, self.ignore_names
+            )
+
+        def has_noqa(lineno, typ):
+            if typ == "property" and sys.version_info >= (3, 8):
+                # python3.8 onwards, lineno for property is the same as that of
+                # the decorated method.
+                lineno = lineno - 1
```
RJ722: Also, I think, it rather makes sense to ignore this case:

```py
class Foo:
    @property
    def bar():  # noqa
        pass
```

than adding `# noqa` infront of `@property`?

tests/test_noqa.py
```diff
@@ -0,0 +1,245 @@
...
+    match = NOQA_REGEXP.search(line)
+    parsed = [
+        c.strip() for c in (match.groupdict()["codes"] or "all").split(",")
+    ]
+    assert parsed == codes
+
+
+@pytest.mark.parametrize(
+    "line",
+    [("# noqa: 123V"), ("# noqa explaination: V012"), ("# noqa: ,V001"),],
```
RJ722: Remove trailing comma.

tests/test_noqa.py
```diff
@@ -0,0 +1,245 @@
...
+        ("# noqa Hi, go on.", ["all"]),
+        ("# noqa: V001", ["V001"]),
+        ("# noqa: V001, V007", ["V001", "V007"]),
+        ("# NoQA: V001,      V003, \t V004", ["V001", "V003", "V004"]),
+    ],
+)
+def test_noqa_regex_present(line, codes):
+    match = NOQA_REGEXP.search(line)
+    parsed = [
+        c.strip() for c in (match.groupdict()["codes"] or "all").split(",")
```
jendrikseipp: Can you factor this out into a function in core.py?

tests/test_report.py
```diff
@@ -39,13 +39,13 @@ def test_report(code, expected, make_whitelist=False):
...
-{filename}:9: unused variable 'foobar' (60% confidence)
-{filename}:11: unreachable code after 'return' (100% confidence)
-{filename}:13: unused property 'myprop' (60% confidence)
+<V004> {filename}:1: unused import 'foo' (90% confidence)
+<V002> {filename}:3: unused class 'Foo' (60% confidence)
+<V003> {filename}:7: unused function 'bar' (60% confidence)
+<V001> {filename}:8: unused attribute 'foobar' (60% confidence)
+<V007> {filename}:9: unused variable 'foobar' (60% confidence)
+<V006> {filename}:11: unreachable code after 'return' (100% confidence)
+<V005> {filename}:13: unused property 'myprop' (60% confidence)
```
jendrikseipp: I think we should mimic the flake8 format more closely (test.py:2:5: F841 local variable 'foo' is assigned to but never used), so for example:

{filename}:1: V004 unused import 'foo' (90% confidence)

tests/test_noqa.py
```diff
@@ -0,0 +1,245 @@
...
+
+class Cellar:  # noqa: V002
+    @property  # noqa: V005
+    def wine(self, grapes):  # noqa: V007
+        pass
+
+    def serve(self, quantity=50):  # noqa: V003
+        self.quantity_served = quantity  # noqa: V001
+        return
+        self.pour()  # noqa: V006
```
jendrikseipp: Shouldn't this be V007?

vulture/core.py
jendrikseipp: I think adding the noqa on the same line as bar() is the better solution, yes. We should change this at a lower level as well: in all Python versions, we should report the line containing bar() as the line number for the property. This should be a separate PR though. However, this might be tricky in case there are multiple @property decorators. Could you try this out, please?

vulture/core.py
```diff
@@ -457,26 +495,41 @@ def _define(
...
+                )
+            )
+
         last_node = last_node or first_node
         typ = collection.typ
-        if (ignore and ignore(self.filename, name)) or _match(
-            name, self.ignore_names
-        ):
+        first_lineno = first_node.lineno
+        if ignored(name) or has_noqa(first_lineno, typ):
```
jendrikseipp: `has_noqa` can become part of `ignored`

vulture/core.py
```diff
@@ -46,6 +47,28 @@
...
 
+ERROR_CODES = {
+    "attribute": "V001",
+    "class": "V002",
+    "function": "V003",
+    "import": "V004",
+    "property": "V005",
+    "unreachable_code": "V006",
+    "variable": "V007",
+}
```
jendrikseipp: I think we should follow pycodestyle and start with error code V101 instead of V001. Since unreachable_code is somewhat different from unused code, I'd give it number V201 (and unused variables V107). (I see now that you changed the order.)

RJ722: If we start with V101, and don't assign unreachable_code in that series, unused_variables would be V106, no?

jendrikseipp: That's right :-)

tests/test_noqa.py
```diff
@@ -0,0 +1,274 @@
...
+def test_noqa_regex_no_groups(line):
+    assert NOQA_REGEXP.search(line).groupdict()["codes"] is None
+
+
+@pytest.mark.parametrize(
+    "line",
+    [("#noqa"), ("##noqa"), ("# n o q a"), ("#NOQA"), ("# Hello, noqa")],
+)
+def test_noqa_regex_not_present(line):
+    assert bool(NOQA_REGEXP.search(line)) is False
```
jendrikseipp: Why not `assert not NOQA_REGEXP.search(line)`?

tests/test_noqa.py
```diff
@@ -0,0 +1,274 @@
...
+    match = NOQA_REGEXP.search(line)
+    parsed = _parse_error_codes(match)
+    assert parsed == codes
+
+
+@pytest.mark.parametrize(
+    "line",
+    [
+        ("# noqa: 123V"),
+        ("# noqa explaination: V012"),
```
jendrikseipp: explaination -> explanation

tests/test_noqa.py
```diff
@@ -0,0 +1,274 @@
...
+        self.pour()  # noqa
+"""
+    )
+    check(v.unused_attrs, [])
+    check(v.unused_classes, [])
+    check(v.unused_funcs, [])
+    check(v.unused_imports, [])
+    check(v.unused_props, [])
+    check(v.unreachable_code, [])
+    check(v.unused_vars, [])
```
jendrikseipp: Can you use exactly one noqa for each category?

tests/test_noqa.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+def test_noqa_specific_issue_codes(v):
+    v.scan(
+        """\
+import this  # noqa: V104
+
+class Cellar:  # noqa: V102
+    @property
+    def wine(self, grapes):  # noqa: V105, V106
```
jendrikseipp: Same here. let's add a separate unused variable inside of this method. The we can also remove the `pass`.

tests/test_noqa.py
```diff
@@ -0,0 +1,274 @@
...
+def test_noqa_functions(v):
+    v.scan(
+        """\
+def play(tune, instrument='bongs', _hz='50'):  # noqa: V103
+    pass
+
+
+# noqa
+def problems():  # noqa: V104
+    ''' They don't go away. :-)'''
```
jendrikseipp: Too pessimistic for my taste :-) I'd remove the comment.

tests/test_noqa.py
```diff
@@ -0,0 +1,274 @@
...
+            a_sheep.grow_hair()  # noqa: V201
+        a_sheep.shave()
+    return
+    for a_sheep in sheep:  # noqa: V201
+        if a_sheep.still_has_hair:
+            a_sheep.shave_again()
+"""
+    )
+    check(v.unreachable_code, [])
+    check(v.unused_funcs, ["shave_sheep"])
```
jendrikseipp: Hehe, nice code :-)

tests/test_report.py
```diff
@@ -39,13 +39,13 @@ def test_report(code, expected, make_whitelist=False):
...
-{filename}:9: unused variable 'foobar' (60% confidence)
-{filename}:11: unreachable code after 'return' (100% confidence)
-{filename}:13: unused property 'myprop' (60% confidence)
+{filename}:1: V104 unused import 'foo' (90% confidence)
+{filename}:3: V102 unused class 'Foo' (60% confidence)
+{filename}:7: V103 unused function 'bar' (60% confidence)
+{filename}:8: V101 unused attribute 'foobar' (60% confidence)
+{filename}:9: V106 unused variable 'foobar' (60% confidence)
+{filename}:11: V201 unreachable code after 'return' (100% confidence)
+{filename}:13: V105 unused property 'myprop' (60% confidence)
```
jendrikseipp: Can you add to the changelog that the output format changes?

vulture/core.py
```diff
@@ -461,26 +501,45 @@ def _define(
...
+            # if name not ignored, check if annotated with "# noqa"
+            if typ == "property" and sys.version_info < (3, 8):
+                # for python < 3.8, increment as many line numbers as there are
+                # decorators on the method.
+                # python3.8 onwards, lineno for property is the same as that of
+                # the decorated method.
+                lineno = lineno + len(first_node.decorator_list)
+            return lineno in (
+                self.noqa_matches[ERROR_CODES[typ]].union(
+                    self.noqa_matches["all"]
```
jendrikseipp: I think it's faster and more readable to test the two sets separately instead of computing the union just for the test.

vulture/core.py
```diff
@@ -461,26 +501,45 @@ def _define(
...
+            ):
+                return True
+
+            # if name not ignored, check if annotated with "# noqa"
+            if typ == "property" and sys.version_info < (3, 8):
+                # for python < 3.8, increment as many line numbers as there are
+                # decorators on the method.
+                # python3.8 onwards, lineno for property is the same as that of
+                # the decorated method.
+                lineno = lineno + len(first_node.decorator_list)
```
jendrikseipp: I think this means that the line numbers we report for unused properties now differ from the line numbers that must have noqa on them. This should be fixed in a separate PR. Can you add a TODO to the TODO file, please?

vulture/core.py
```diff
@@ -461,26 +501,45 @@ def _define(
...
+                name, self.ignore_names
+            ):
+                return True
+
+            # if name not ignored, check if annotated with "# noqa"
+            if typ == "property" and sys.version_info < (3, 8):
+                # for python < 3.8, increment as many line numbers as there are
+                # decorators on the method.
+                # python3.8 onwards, lineno for property is the same as that of
+                # the decorated method.
```
jendrikseipp: Please start the first sentence with an uppercase "For" and the second with "From".

tests/test_noqa.py
RJ722: I just didn't happen to think of it. ':D Changing it.

tests/test_noqa.py
RJ722: Do you mean, break it down into individual tests for all of them?

vulture/core.py
RJ722: I raised #200 :D -- I would be happy to rebase this branch, once we get it merged.

vulture/core.py
RJ722: That might result in the same or v/s union bug, I guess. Do you have something in mind?

tests/test_noqa.py
RJ722: I added the variable - `year`. Just making sure that you in fact, meant this. If yes, can we mark this as resolved?

tests/test_noqa.py
jendrikseipp: No, I was just surprised to see 6 noqas but 7 checks. Adding the `year` variable here as well should resolve this.

tests/test_noqa.py
jendrikseipp: I'd drop the useless return and remove the V106 noqa from the line above, but then it's resolved, yes.

vulture/core.py
jendrikseipp: `return lineno in self.noqa_matches[ERROR_CODES[typ]] or lineno in self.noqa_matches["all"]`

RJ722: Yep, thanks!

tests/test_noqa.py
RJ722: Ahh, got it.

tests/test_noqa.py
RJ722: Aah, okay. I took care of this.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Unreleased
 
 * Ignore unused import statements which occur in `__init__.py` (RJ722, #192).
+* Report issue codes in ouput (e.g., `code.py:1: V104 unused import ...`)
```
jendrikseipp: ouput -> output

vulture/core.py
```diff
@@ -55,6 +78,13 @@ def _get_unused_items(defined_items, used_names):
     return unused_items
 
 
+def _parse_error_codes(match):
+    # if no code is specified, assign it to `all`
```
jendrikseipp: Start with upper case and end with full stop. I#d also be a little more concrete: 

If no error code is specified, add the line to the "all" category.

vulture/core.py
```diff
@@ -461,26 +501,44 @@ def _define(
         confidence=DEFAULT_CONFIDENCE,
         ignore=None,
     ):
+        def ignored(lineno):
+            if (ignore and ignore(self.filename, name)) or _match(
+                name, self.ignore_names
+            ):
+                return True
+
+            # if name not ignored, check if annotated with "# noqa"
```
jendrikseipp: Rephrase: Check if the reported line is annotated with "# noqa".

RJ722:
Now, that #200 has been merged (yay! \o/), let me rebase this branch and resolve conflicts.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 * Ignore unused import statements which occur in `__init__.py` (RJ722, #192).
 * Report first decorator's line number for decorated objects on Python 3.8+.
   (RJ722, #200)
+* Report issue codes in output (e.g., `code.py:1: V104 unused import ...`)
+  (RJ722, #195)
+* Support `# noqa` comments to suppress results on that line. (RJ722, #195).
```
jendrikseipp: Could you please draft the documentation for the README.md file? It's always easier to edit existing text than writing new text.

vulture/core.py
```diff
@@ -205,14 +236,23 @@ def get_set(typ):
 
         self.ignore_names = ignore_names or []
         self.ignore_decorators = ignore_decorators or []
+        self.noqa_matches = defaultdict(set)
 
         self.filename = ""
         self.code = []
         self.found_dead_code_or_error = False
 
+    def parse_noqa(self, code):
```
jendrikseipp: It's a bit strange that the method is passed `self.code`, but not `self.noqa_matches` even though both are instance attributes. I think we should pass both variables to the method and turn it into a function. Then we can collect all code related to noqa in a vulture/noqa.py file, i.e., parse_noqa(), _parse_error_codes() and NOQA_REGEXP.

vulture/core.py
```diff
@@ -461,14 +501,25 @@ def _define(
...
+            if (ignore and ignore(self.filename, name)) or _match(
+                name, self.ignore_names
+            ):
+                return True
+
+            # Check if the reported line is annotated with "# noqa".
+            return (
+                lineno in self.noqa_matches[ERROR_CODES[typ]]
+                or lineno in self.noqa_matches["all"]
+            )
```
jendrikseipp: This can become a function `noqa.ignore_line(noqa_matches, lineno, typ)` and then we can add it to the if-statement above.

vulture/core.py
```diff
@@ -205,14 +236,23 @@ def get_set(typ):
 
         self.ignore_names = ignore_names or []
         self.ignore_decorators = ignore_decorators or []
+        self.noqa_matches = defaultdict(set)
```
jendrikseipp: Please rename noqa_matches to noqa_lines everywhere.

CHANGELOG.md
RJ722: Sure!

jendrikseipp:
Thanks for the additional changes! I edited README.md slightly. Could you please resolve the merge conflicts? Then I can merge the PR.

RJ722:
@jendrikseipp, thanks for your timely reviews, and help on this! :-)

PR #200, opened 2020-03-14
Report method definition lineno when reporting property

RJ722:
## Description

Previously, we reported the lineno on which the `@property` annotation was made. It, however makes sense to report the line on which the method is actually defined.

This was also non-consistent behavior b/w Vulture running on Python < 3.8 and >= 3.8.

## Related Issue

https://github.com/jendrikseipp/vulture/pull/195#discussion_r392578798

## Checklist:


RJ722:
@jendrikseipp I took the liberty of adding some more simple, but more concrete tests to make sure that we do check for properties of individual unused items somewhere. Let me know if this doesn't right.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 
 * Ignore unused import statements which occur in `__init__.py` (RJ722, #192).
+* When reporting properties, use lineno on which the corresponding method is
```
jendrikseipp: "Report first line number of method definition for unused properties (RJ722, #200)."

RJ722: Done.

vulture/core.py
```diff
@@ -461,14 +461,23 @@ def _define(
         confidence=DEFAULT_CONFIDENCE,
         ignore=None,
     ):
+        def _get_first_lineno(first_node, typ):
+            if typ == "property" and sys.version_info < (3, 8):
+                # For python < 3.8, increment as many line numbers as there are
+                # decorators on the method.
+                # From Python3.8 onwards, lineno for property is the same as
+                # that of decorated method.
```
jendrikseipp: Let's turn this into a docstring and rewrite it slightly:

For decorated methods, Python < 3.8 uses the first decorator to report the first line number. To obtain the line number of the actual method definition (as in Python >= 3.8), we add the number of decorators to the line number. Obviously, this is only exact if all decorators use only one line.

jendrikseipp: Can we maybe make this an exact line number (even for multiline decorators)? For this, we would need to be able to retrieve the method's AST node from the property's AST node and call `.lineno` on it. Does the property's AST node have the relevant fields? You can inspect this on Python 2.7 with `dir(node)`.

RJ722: > Obviously, this is only exact if all decorators use only one line.

Yeah, that is a problem. We need more test cases here.

> For this, we would need to be able to retrieve the method's AST node from the property's AST node and call .lineno on it

I'm a little confused with this. Aren't AST nodes for property and method the same?

jendrikseipp: What I meant was the FunctionDef AST node and a *decorator* AST node. To retrieve the method line number from the FunctionDef node `node`, we could use `lines.get_last_line_number(node.decorator_list[-1])`. However, this is also just an approximation. Thinking about this again, I belive that it would be best to simply report the first line number of the first decorator for unused properties for Python 3.8+. This preserves the behaviour of Vulture for newer Python versions and is also more accurate for counting the size of the unused code chunk (if the property is unused, we also don't need it's decorators). Most importantly, it's easy to compute the line number accurately (should work for all Python versions, but deserves a comment about why we don't use `node.lineno`): `node.decorator_list[0].lineno`.

RJ722: I tried `lines.get_last_line_number(node.decorator_list[-1])` for a bunch of cases, and I agree about the inaccuracy, eg., it would fail here:

```python
class TestFoo:
    @pytest.mark.parametrize(
        "line, codes",
        [
            ("# noqa", ["all"]),
            ("## noqa", ["all"]),
        ],
    )
    @property
    @pytest.mark.parametrize(
        "line, codes",
        [
            ("# noqa", ["all"]),
        ],
    )
    def test_noqa_regex_present(line, codes):
        pass
```

Also, I can't remember why this PR only targeted properties, and not all decorated functions? 😅

Anyways, I'll update this PR for all functions with the changes you mentioned above.

RJ722:
Apologies for the delay @jendrikseipp. I've updated the PR according to the recent discussion.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 
 * Ignore unused import statements which occur in `__init__.py` (RJ722, #192).
+* Report first decorator's line number for decorated objects on Python 3.8+.
+  (RJ722, #200)
```
RJ722: I'm not sure -- do we need to create an entry in CHANGELOG for this?

vulture/lines.py
```diff
@@ -63,3 +64,19 @@ def get_last_line_number(node):
...
+    if typ in decorated_typs and sys.version_info >= (3, 8):
+        # From Python3.8 onwards, lineno for decorated objects is the
+        # line at which the object definition starts, which is
+        # different from what Python < 3.8 did -- reporting the lineno
+        # of the first decorator. To preserve this behaviour of Vulture
+        # for newer Python versions, which is also more accurate for
+        # counting the size of the unused code chunk (if the property
+        # is unused, we also don't need it's decorators), we return the
+        # lineno of the first decorator, if there are any.
+        if first_node.decorator_list:
```
jendrikseipp: Let's rewrite this slightly: 

```
if sys.version_info >= (3, 8):
    decorators = getattr(node, "decorator_list", [])
    if decorators:
        return decorators[0].lineno

```

vulture/lines.py
```diff
@@ -63,3 +64,19 @@ def get_last_line_number(node):
             except AttributeError:
                 pass
         node = last_child
+
+
+def get_first_line_number(first_node, typ):
+    decorated_typs = ("class", "function", "property")
+    if typ in decorated_typs and sys.version_info >= (3, 8):
+        # From Python3.8 onwards, lineno for decorated objects is the
```
jendrikseipp: Python3.8 -> Python 3.8

vulture/lines.py
```diff
@@ -63,3 +64,19 @@ def get_last_line_number(node):
...
+        # line at which the object definition starts, which is
+        # different from what Python < 3.8 did -- reporting the lineno
+        # of the first decorator. To preserve this behaviour of Vulture
+        # for newer Python versions, which is also more accurate for
+        # counting the size of the unused code chunk (if the property
+        # is unused, we also don't need it's decorators), we return the
+        # lineno of the first decorator, if there are any.
```
jendrikseipp: Nice text!

vulture/lines.py
```diff
@@ -63,3 +64,19 @@ def get_last_line_number(node):
             except AttributeError:
                 pass
         node = last_child
+
+
+def get_first_line_number(first_node, typ):
```
jendrikseipp: first_node -> node

vulture/lines.py
RJ722: Hehe, I just copied over your comments.

jendrikseipp:
Thanks for your work and persistence!

PR #201, opened 2020-04-09
Proposal: Enable DeepSource for detecting issues

RJ722:
## Description

Hey @jendrikseipp,

I ran [DeepSource][ds] on Vulture, and found several interesting issues - https://deepsource.io/gh/RJ722/vulture/issues/ - some of which I think we should fix. (Let's also use this PR to triage these issues)

Also, I think that it would be helpful to have DeepSource enabled here on the main repos, so that we get to see analysis results for all new PRs, and catch issues early on.

Also, a disclaimer: I am currently a part of the DeepSource team, working on their analyzers.

[ds]: https://deepsource.io/

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,14 @@
+version = 1
+
+test_patterns = [
+  "tests/*.py",
+]
+
+exclude_patterns = [
+  "vulture/whitelists/*",
+]
+
+[[analyzers]]
+name = "python"
+enabled = true
+runtime_version = "3.x.x"
--- a/dev/make-release-notes.py
+++ b/dev/make-release-notes.py
@@ -3,7 +3,7 @@
 import re
 import sys
 
-_, VERSION, CHANGELOG, LIST = sys.argv
+_, VERSION, CHANGELOG, LIST = sys.argv  # skipcq: PYL-W0632
 HEADER_REGEX = fr"# {VERSION} \(\d\d\d\d-\d\d-\d\d\)\n"
 
 notes_list = []
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -112,7 +112,7 @@ def _ignore_variable(filename, varname):
     )
 
 
-class Item(object):
+class Item(object):  # skipcq: PYL-R0205
     """
     Hold the name, type and location of defined code.
     """
@@ -270,6 +270,7 @@ class Vulture(ast.NodeVisitor):
             self._log("Scanning:", module)
             try:
                 module_string = utils.read_file(module)
+            # skipcq: PYL-W0641
             except utils.VultureInputException as err:  # noqa: F841
                 print(
                     "Error: Could not read file {module} - {err}\n"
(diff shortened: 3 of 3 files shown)


jendrikseipp:
Thanks for looking into this! The DeepSource report is interesting, but only contains a few issues that I'd actually like to fix:

https://deepsource.io/gh/RJ722/vulture/issue/PYL-E0101/occurrences
https://deepsource.io/gh/RJ722/vulture/issue/PYL-W0613/occurrences
https://deepsource.io/gh/RJ722/vulture/issue/PYL-W0201/occurrences

I think these should be tackled in a separate PR.

Regarding adding DeepSource to Vulture's CI, I doubt that it's worth the trouble, since we would need to maintain a long list of false positives. But feel free to compile this list and let's see how complicated it is.

RJ722:
Thanks for traiging the issues! Apart from the ones you mentioned, I think that we should also fix the following:

- https://deepsource.io/gh/RJ722/vulture/issue/BAN-B101/occurrences
- https://deepsource.io/gh/RJ722/vulture/issue/PYL-R1705/occurrences

What do you think?

Also, I'd open individual GitHub issues for each one of these, which could be then publicised as "Good First Issues".

> But feel free to compile this list and let's see how complicated it is.

Thanks! I've added the `# skipcq` comments in this PR itself.

Other than this, there isn't much maintainence overhead either - I added the following issues to be ignored repository wide (single button click) -- found them too stringent for my taste:

* One line docstring should fit on one line with quotes.
* Missing docstring in public function
* Audit required: Detected subprocess call
* Import of `subprocess` module detected.

Screen Shot 2020-04-10 at 11 05 58 PM

jendrikseipp:
I think the assert statement is fine. It will catch programming errors when running things unoptimized locally. The "unnecessary else" case is a question of style preference. Personally, I often like the symmetry of the if-else structure.

> Also, I'd open individual GitHub issues for each one of these, which could be then publicised as "Good First Issues".

I appreciate the idea, but I think fixing these directly would require much less work.

> Thanks! I've added the # skipcq comments in this PR itself.

I think these make the code harder to read. I think the DeepSource results contain too many false positives for the system to be run in CI. In other words, it's interesting to see these results, but I don't think it's a good benefit/work ratio to add them to our permanent test suite. We may want to revisit this once DeepSource supports more interesting tests and produces fewer false positives.

RJ722:
> > Also, I'd open individual GitHub issues for each one of these, which could be then publicised as "Good First Issues".

> I appreciate the idea, but I think fixing these directly would require much less work.

Okay, let me fix them in a separate PR.

> I think these make the code harder to read. I think the DeepSource results contain too many false positives for the system to be run in CI. In other words, it's interesting to see these results, but I don't think it's a good benefit/work ratio to add them to our permanent test suite. We may want to revisit this once DeepSource supports more interesting tests and produces fewer false positives.

Thanks for the feedback! :)

PR #213, opened 2020-07-08
add trove classifier for py38

ju-sh:
## Description
Add trove classifier for Python3.8

Add trove classifier for Python3.8. Thought it was too small a change to be creating an issue.

As an aside, can't we open the files, read its contents, assign it to a variable and use that variable in the `long_description` argument to `setup()` in setup.py instead of opening it then and there?

## Checklist:

- [] My change requires a change to the documentation in the README file.
- [] I have updated the documentation accordingly.
- [] I have added an entry in CHANGELOG.md.
- [] I have added tests to cover my changes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,6 +48,7 @@ setuptools.setup(
         "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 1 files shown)


jendrikseipp:
Thanks!

> As an aside, can't we open the files, read its contents, assign it to a variable and use that variable in the `long_description` argument to `setup()` in setup.py instead of opening it then and there?

Yes, I guess we could. Feel free to make the change.

Issue #164, opened 2019-07-12
Support loading options from a configuration file (pyproject.toml?)


Varriount:
It would be nice if vulture supported using setup.cfg/setup.ini to retrieve whitelists, exclusions, etc. from.

RJ722:
Hi @Varriount,

Thanks for the feedback, but I'm not sure that I entirely get what you're trying to say. Can you please give an example of what exactly are you asking for, please?

tarkatronic:
Alternatively, perhaps `pyproject.toml` support would be good, as this is becoming The Way for configuration in the Python world. An example entry for `vulture` might look like:
```toml
[tool.vulture]
min-confidence = 70
exclude = ".eggs,.git,.*cache,.tox,.venv,build,dist"
```
This would allow a user to have more determinism in their `vulture` runs, since they would not have to specify command line options every time.

I recently implemented this in [tartufo](https://github.com/godaddy/tartufo/blob/ac18f615f0c2f7ae59813790cefcbce4214d7958/tartufo/config.py#L24..L51), based heavily on the work in [black](https://github.com/psf/black/blob/de6f4b1f6bd3211e80bc5a5f1f88e6ef69a550aa/black.py#L216..L249).

This style of configuration is being adopted in many other large Python tools. A list of other tools can be found here: https://github.com/flying-sheep/awesome-python-packaging

In addition, [mypy](https://github.com/python/mypy/issues/5205) will be adding support as well.

jendrikseipp:
@tarkatronic sounds interesting! It would be awesome if you could make a pull request.

RJ722:
@jendrikseipp I can work on this. However, adding support for `pyproject.toml` would mean having an external dependency `toml`. Are you okay with it?

jendrikseipp:
Good point. I'm not sure. Having no external dependencies certainly has appeal. Also, it's not clear which TOML library we should use (there are 3 Python TOML libraries). Maybe we should defer this issue until the dust has settled on how to declare options for linters and similar tools.

jendrikseipp:
I guess we could make the TOML dependency optional like isort does. Still, it's not clear which TOML library to use.

RJ722:
https://pypi.org/project/toml/ seems to be the standard and most popular (more than 18k users according to https://github.com/uiri/toml)

jendrikseipp:
I checked which projects use which toml library:
* toml: black, isort, mypy (WIP)
* tomlkit: poetry

[...]

PR #215, opened 2020-07-12
Implement support for a pyproject.toml config file

exhuma:
## Description
This change will read config values from a `pyproject.toml` file first, and then update those values with the CLI arguments.

## Related Issue

See #164 

## Checklist:


exhuma:
I just noticed the failure on Windows. I'll check this once I have access to a windows box. I assume I'll just have to use an older version of the toml lib.

exhuma:
Okay.... so it seems `toml` is not available for pypy on Windows. Question: Considering that `vulture` is CLI-tool, aiding in development of other applications, is it really necessary to support `pypy`? It's an honest question. I don't mind investing more time in making this work on pypy+win as well (if really necessary).

jendrikseipp:
I think this problem is caused by the Github Action using different installation options on Windows compared to Linux and MacOS (see .github/workflows/main.yml). My guess is we only need to remove the --no-index and --find-links arguments. The current command works on Windows because Vulture didn't have dependencies. I think @RJ722 wanted to look into this anyway, so feel free to ignore this for this pull request. But of course you can also try to fix the GitHub Action.

exhuma:
That makes sense. Tomorrow I'll be out of the week-end and likely programming again all day, and after a whole day of coding my batteries are usually running on empty. I'll see if I can squeeze some time in for this. Sounds like a trivial fix. Whoever gets to it first then I guess 😉

exhuma:
Oh... in the PR I didn't add any design decisions I took along the way. The important ones can be summarised as follows:

* I'm assuming `vulture` will be run from the project root as CWD when looking for `pyproject.toml`. This may not be as flexible as it could be, but it keeps the code-base *way* simpler. `black` does a fairly complex lookup, and I don't think that this added code-complexity is worth the gain. `flit` on the other hand also assumes it will be run from the project root. I decided to go the "flit" way.
* The argument parser was moved to the new `config.py` file to keep everything related together in one module.
* I added type-hints because they don't hurt, and I've gotten used to them. Especially for nicer editor autocompletion.
* I decided to subclass `dict` for the configobject to ensure backwards-compatibility with existing code, while still being able to leverage `dict.update()`. I started off with a slightly more cumbersome aproach first with a data-class (albeit not a "real" dataclass due to the Python-2 dependency). This would have given rich type-hinting, but would have introduced cumbersome code for "merging" the two configs (argparse & toml). So I opted for the `dict` aproach.

tests/test_config.py
```diff
@@ -0,0 +1,72 @@
...
+        min_confidence = 10
+        sort_by_size = true
+        verbose = true
+        paths = ['path1', 'path2']
+        """
+        )
+    )
+    result = _parse_toml(data)
+    assert isinstance(result, Config)
+    assert result == expected
```
jendrikseipp, line 72: Please end all sentences with dots.

vulture/config.py
```diff
@@ -0,0 +1,168 @@
...
+from os.path import exists
+from typing import Any, Dict, List, Optional
+from typing.io import TextIO
+
+import toml
+
+from .version import __version__
+
+MIN_CONFIDENCE_DEFAULT = 0
+
```
jendrikseipp, line 16: I think we should add a list or dict of option names and their defaults. Then we can check that the TOML file has no extraneous keys and can loop over the list when creating Config objects instead of listing all option names twice.

vulture/config.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+    parser.add_argument(
+        "--sort-by-size",
+        action="store_true",
+        help="Sort unused functions and classes by their lines of code.",
+    )
+    parser.add_argument("-v", "--verbose", action="store_true")
+    parser.add_argument("--version", action="version", version=version)
+    namespace = parser.parse_args(args)
+    output = Config(
```
jendrikseipp, line 127: No need for variable `output` (here and above).

vulture/config.py
```diff
@@ -0,0 +1,168 @@
...
+    # type: () -> Config
+    """
+    Returns a config object for vulture, merging both ``pyproject.toml`` and
+    CLI arguments (CLI will have precedence).
+    """
+    if exists("pyproject.toml"):
+        with open("pyproject.toml") as toml_config:
+            app_config = _parse_toml(toml_config)
+    cli_config = _parse_args()
+    app_config.update(cli_config)
```
jendrikseipp, line 167: I think the CLI defaults will override all TOML settings, won't they? In any case, we should add a test for this.

vulture/config.py
jendrikseipp: It would be good to specify the defaults only once, otherwise they might go out of sync. But maybe we also only need to specify them once, because the CLI options override the TOML options anyway.

README.md
```diff
@@ -106,6 +106,24 @@ You can use the `--min-confidence` flag to set the minimum confidence
 for code to be reported as unused. Use `--min-confidence 100` to only
 report code that is guaranteed to be unused within the analyzed files.
 
+## Configuration
+
+CLI arguments can also be added to `pyproject.toml` into the `tool.vulture`
+section. Leading dashes are removed and remaining dashes are replaced with
+underscores. Example:
```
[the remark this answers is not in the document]
exhuma: Addressed in edad0e4

README.md
```diff
@@ -106,6 +106,24 @@ You can use the `--min-confidence` flag to set the minimum confidence
...
+``` toml
+[tool.vulture]
+exclude = ['exclude1', 'exclude2']
+ignore_decorators = ['deco1', 'deco2']
+ignore_names = ['name1', 'name2']
+make_whitelist = true
+min_confidence = 10
+sort_by_size = true
+verbose = true
+paths = ['path1', 'path2']
```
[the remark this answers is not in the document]
exhuma: Addressed in 9202909

README.md
```diff
@@ -106,6 +106,24 @@ You can use the `--min-confidence` flag to set the minimum confidence
...
 
+## Configuration
+
+CLI arguments can also be added to `pyproject.toml` into the `tool.vulture`
+section. Leading dashes are removed and remaining dashes are replaced with
+underscores. Example:
+
+``` toml
+[tool.vulture]
+exclude = ['exclude1', 'exclude2']
```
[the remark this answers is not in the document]
exhuma: Addressed in 8888997

vulture/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "1.5"
```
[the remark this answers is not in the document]
exhuma: Addressed in a60ee5d

vulture/config.py
```diff
@@ -0,0 +1,168 @@
...
+        return super(Config, self).__getattribute__(name)
+
+
+def make_config():
+    # type: () -> Config
+    """
+    Returns a config object for vulture, merging both ``pyproject.toml`` and
+    CLI arguments (CLI will have precedence).
+    """
+    if exists("pyproject.toml"):
```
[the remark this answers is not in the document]
exhuma: Addressed in 7d86cf9

vulture/config.py
```diff
@@ -0,0 +1,168 @@
...
+    return output
+
+
+def _parse_args(args=None):
+    # type: (Optional[List[str]]) -> Config
+    """
+    Parse CLI arguments
+
+    :param args: A list of strings representing the CLI arguments. If left to
+        the default, this will default to sys.argv
```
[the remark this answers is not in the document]
exhuma: Addressed in 1a20844

tests/test_config.py
exhuma: Addressed in 1a20844 ... I guess :wink:

tox.ini
```diff
@@ -14,6 +14,8 @@ deps =
   coverage
   pytest
   pytest-cov
+  toml
```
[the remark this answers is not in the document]
exhuma: Yes, you're right. Addressed in 96fa992

vulture/config.py
exhuma: Addressed in fc7d3fa

vulture/config.py
exhuma: Without this line, the CLI defaults will not automatically override them (unless I'm missing something). I agree about the test, although I need to refactor some things for simple DI (because of the `open()` call).

exhuma: Addressed in ae44f09

vulture/config.py
exhuma: Addressed in 9b19a1a

exhuma:
As you mentioned in a comment above, I have removed the `--no-index` flag from `pip install`. But I wonder if there was a specific reason for adding that in the first place. Might be worthwile having @RJ722 signing it off.

exhuma:
After a little shower-thought this morning I added a final change to the config object: Following the [standard docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__) it should either return a valid expression or something enclosed in `<...>`. Being a subclass of `dict`, the repr returned a valid expression, but that expression would not result in an equivalent instance.

So I added these two lines making this so.

It has the added benefit of making any debugging/logging lines more explicit as well if they include the config object.

vulture/whitelists/config_whitelist.py
```diff
@@ -0,0 +1,11 @@
...
+from typing.io import TextIO
+from vulture import config
+
+List
+Optional
+TextIO
+
+# NOTE: This was added to make unit-tests pass and allow commiting intermediate
+# work. It should be removed once the TOML parsing is fully implemented.
+config._parse_toml
```
jendrikseipp, line 11: Do we still need this file? I guess you just haven't removed all traces of the `typing` code, yet.

tests/test_config.py
```diff
@@ -0,0 +1,116 @@
...
+        paths=["cli_path"],
+        exclude=["cli_exclude"],
+        ignore_decorators=["cli_deco"],
+        ignore_names=["cli_name"],
+        make_whitelist=True,
+        min_confidence=20,
+        sort_by_size=True,
+        verbose=True,
+    )
+    assert result == expected
```
jendrikseipp, line 116: Thanks for adding the test. It works, because all values are set in both CLI and TOML. However, I think it will fail if I add `verbose = true` to the TOML file, but don't pass `--verbose` on the command line. We need to test this scenario, in this test, or a separate test, whatever you prefer.

vulture/config.py
```diff
@@ -0,0 +1,184 @@
...
+        "verbose": False,
+    }
+
+    def __getattribute__(self, name):
+        if name in self:
+            return self[name]
+        return super(Config, self).__getattribute__(name)
+
+    def __repr__(self):
+        return "Config(%r)" % dict(self)
```
jendrikseipp, line 40: Can you use str.format() for new code please?

vulture/config.py
```diff
@@ -0,0 +1,184 @@
...
+        "ignore_names": [],
+        "make_whitelist": False,
+        "sort_by_size": False,
+        "verbose": False,
+    }
+
+    def __getattribute__(self, name):
+        if name in self:
+            return self[name]
+        return super(Config, self).__getattribute__(name)
```
jendrikseipp, line 37: I wonder whether the code wouldn't be simpler if we didn't subclass dict, but used dict directly? What do you think? I think the only thing to change would be to turn the staticmethod into a function that returns a dict.

RJ722:
Thanks for the PR @exhuma! I am looking into the Windows wheel problem right now: I can't seem to recall why `--no-index` was added, but if it works without that, hooray! \o/

Though, I'm trying to make the builds a little more cleaner, by unifying the way we install wheels on all the platforms. One of the problems I can remember is that windows doesn't support wildcard/glob pattern matching, so I cannot directly write `pip install dist/*.whl`. I'll explore it more, run a bunch of experiments on my fork and raise a PR. However, let's not keep that PR a blocker on this.

The-Compiler:
Out of curiosity, why do you build/install wheels on CI (rather than using e.g. tox) on CI at all?

jendrikseipp:
> Out of curiosity, why do you build/install wheels on CI (rather than using e.g. tox) at all?

I think we wanted to test that building wheels works (especially on Windows). From my understanding, pip would silently fall back to installing from source if building the wheel fails. But maybe this belongs in @RJ722's upcoming pull request. I'd appreciate your feedback there on how to simplify the tests.

exhuma:
Concerning the latest changes, I just wanted to give a sign of life. I'm still here, but during normal office-weeks I don't get all that much time to work on this. I'll try to get it done on Sunday if I don't get around to it before. In any case, I'm with you with the requested changes. I already felt a bit iffy with the dict-subclassing, the rest is housekeeping so I don't have anything substantial to add to the comments :wink:

jendrikseipp:
Sounds good, thanks for the update!

vulture/config.py
exhuma: Addressed this in 06ebedd

exhuma: I agree about this. As mentioned I felt iffy about this as well. And additionally the overriding of `__getattribute__` would break things like `foo.keys()`.

Originally I added this as a poor-man's implementation of a data-class for type-safety, but as typing is not yet in vulture, and because we need to support Python <= 3.5 this is a no-go.

Removed the subclass :)

vulture/whitelists/config_whitelist.py
exhuma: You're right. Addressed in 06ebedd

vulture/config.py
exhuma: Addressed in 9811bb4

tests/test_config.py
exhuma: Should be fine now. See ff30697

The is a slight issue with merging these values which is why I was forced to use a sentinel value to detect unset values. Otherwise the boolean checks would be difficult to write (if not impossible).

Let me know what you think.

exhuma:
Should be good now. I left the review comments as "unresolved" for now. If you're okay with them if would be nice to mark them as resolved. It makes it also easier for me to spot things which need additional work ;)

tox.ini
```diff
@@ -14,6 +14,7 @@ deps =
   coverage
   pytest
   pytest-cov
+  typing; python_version < "3.0"
```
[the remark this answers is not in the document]
exhuma: I find type-hints already quite usable since Python 3.4 (I have been using the extensively ever since). But I'm not hung up on them. Especially as vulture is a CLI tool rather than a library. IMO they make most sense on the external "contact point" of an API/library. Removed in f272778.

vulture/whitelists/config_whitelist.py
```diff
@@ -0,0 +1,11 @@
...
+from typing.io import TextIO
+from vulture import config
+
+List
+Optional
+TextIO
+
+# NOTE: This was added to make unit-tests pass and allow commiting intermediate
+# work. It should be removed once the TOML parsing is fully implemented.
+config.make_config
```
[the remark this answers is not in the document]
exhuma: Removed in f272778

tests/test_config.py
exhuma: Using a sentinel like is a habit of mine, which I use in these situations. Mainly for consistency because sometimes you want to also flag something as having "no default set" while still being able to store the value "None". And if I use `None` to flag something as having "no default" I can no longer determine whether the value was supposed to be `None` or whether I should use the default value. It also keeps things more explicit in code. Reading something like `if valus is NO_DEFAULT` communicates the intent better than reading `if value is None` (IMO). 

Considering that all the values here are booleans, we could use `None` as this special marker.

Let me know what you think. If I will read an imperative sentence in the discussion, I will follow the order :smile:

vulture/config.py
```diff
@@ -0,0 +1,193 @@
...
+    Returns a config object for vulture, merging both ``pyproject.toml`` and
+    CLI arguments (CLI will have precedence).
+
+    :param argv: The CLI arguments to be parsed. This value is transparently
+        passed through to :py:meth:`argparse.ArgumentParser.parse_args`
+    :param tomlfile: An IO instance containing TOML data. By default this will
+        auto-detect an existing ``pyproject.toml`` file and exists solely for
+        unit-testing.
+    """
+    cli_config = _parse_args(argv)
```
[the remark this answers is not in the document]
exhuma: Agreed. Done in 6797dca

vulture/config.py
```diff
@@ -0,0 +1,193 @@
...
+    # consideration.
+    for key in toml_config.keys():
+        cli_value = cli_config.get(key, NO_DEFAULT)
+        if cli_value is not NO_DEFAULT:
+            toml_config[key] = cli_value
+
+    if detected_toml_path and toml_config["verbose"]:
+        print("Reading config values from {}".format(detected_toml_path))
+
+    return toml_config
```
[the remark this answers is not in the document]
exhuma: Agreed. Done in 6e86519

vulture/config.py
```diff
@@ -0,0 +1,193 @@
...
+    remaining_keys = set(data.keys())
+
+    output = {}
+    for key, default in DEFAULTS.items():
+        output[key] = data.get(key, default)
+        remaining_keys.discard(key)
+    for remainder in sorted(remaining_keys):
+        print(
+            "Unprocessed config option {}".format(remainder), file=sys.stderr
+        )
```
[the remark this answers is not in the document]
exhuma: Done in 3a5802d

I feel a bit uncomfortable that I had to use `mock.patch` in the tests for this, but using DI here would make the code a bit more complicated (not overly so though), which is why I leaned towards monkeypatching. Let me know what you think about this.

Another point where I may have gone overboard is the unit-test for `sys.exit`. As it stands, the *test* is fairly tightly coupled to the implementation details of `sys.exit`. But I doubt that such a low-level and important method will ever change in Python. Considering how many OS-level system utilities are written in Python.

The reason the assertion of `sys.exit` call looks a bit complicated is because I aimed for a future-proof test. No matter how `sys.exit` will be called in the future, this assertion ensures that we never exit with a zero exit-code. Thoughts?

exhuma:
Hmmm... the Python-2 tests fail in the CI pipeline, but running `tox -e py27` (and simply `tox`) works on my machine, including Python 2.7.

I addressed the test-regression of 3a5802d in d4e1543, but the CI pipeline still fails. What am I missing?

The-Compiler:
It fails with:

```
tests/test_config.py:6: in <module>
    from unittest.mock import patch
E   ImportError: No module named mock
```

on Python 2.7. That module has only been added to the stdlib in Python 3.3. Not sure why it would work with Python 2.7 on your machine to be honest (though I haven't tried locally).

@jendrikseipp @RJ722 Given that it's 2020, why not drop Python 2.7 support entirely?

exhuma:
@The-Compiler : I added the dependency [in tox.ini](https://github.com/jendrikseipp/vulture/pull/215/commits/d4e1543835562c5a955ec1c8a126b360719c1ae5#diff-b91f3d5bd63fcd17221b267e851608e8R17) And on my local machine it works as expected (using tox, in Python2.7). I don't understand why this is not pulled in by the pipeline.

The-Compiler:
Ah, sorry, I missed that bit. My first guess would be that virtualenv (and its included `pip`) is older on CI and somehow doesn't understand the environment marker correctly. Maybe adding a `pip install --upgrade virtualenv` [around here](https://github.com/jendrikseipp/vulture/blob/master/.github/workflows/main.yml#L32-L35) would help?

jendrikseipp:
> @jendrikseipp @RJ722 Given that it's 2020, why not drop Python 2.7 support entirely?

Yes, I think we should do this soon. I'd rather not do it in this issue though, of course.

exhuma:
Tests are passing now even in Python 2. I left the dependency in `tox.ini` though. If someone clones the repo and runs tox it needs to be in there. So there's a tiny duplication regarding the `mock` dependency.

But if there's plans to drop Python 2 anyway, this shouldn't be too much of an issue. It's fairly easy to drop this from the workflow once Python 2 is gone.

The only remaining thing I see is the discussion about the sentinel value. FWIW, this is a pattern I nicked [from the stdlib](https://github.com/python/cpython/blob/master/Lib/configparser.py#L357).

tests/test_config.py
jendrikseipp: Point taken. Let's keep the special sentinel value :-)

vulture/config.py
```diff
@@ -0,0 +1,194 @@
...
+    output = {}
+    for key, default in DEFAULTS.items():
+        output[key] = data.get(key, default)
+        remaining_keys.discard(key)
+    for remainder in sorted(remaining_keys):
+        print(
+            u"Unprocessed config option {}".format(remainder), file=sys.stderr
+        )
+    if remaining_keys:
+        sys.exit(u"Invalid config!")
```
jendrikseipp, line 48: I think it would be cleaner to rename remaining_keys to unknown_keys and compute them with `unknown_keys = set(data) - set(DEFAULTS)` and then

    if unknown_keys:
        sys.exit("Unknown configuration keys: {}".format(sorted(unknown_keys)))

exhuma:
Again, just a comment to show a sign of life. I won't be able to work on this this week. I'm getting ready to take some time off next week, so I'll be a bit silent for the coming days. But I'll get back to this as soon as possible.

jendrikseipp:
Thanks for the notice!

tests/test_config.py
```diff
@@ -0,0 +1,180 @@
...
+    exit-code
+    """
+    with patch("vulture.config.sys") as msys:
+        from_dict({"unknown_key_1": 1})
+
+    assert len(msys.exit.mock_calls) == 1
+    exit_arg = msys.exit.mock_calls[0].args[0][0]
+    assert (isinstance(exit_arg[0], int) and exit_arg[0] != 0) or isinstance(
+        exit_arg[0], strtype
+    )
```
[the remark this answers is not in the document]
exhuma: Ah, I was unaware of that pytest feature. I have addressed it in c5b9fa9

tests/test_config.py
```diff
@@ -0,0 +1,180 @@
...
+        "cli_path",
+    ]
+    result = make_config(cliargs, toml)
+    assert result["verbose"] is True
+
+
+def test_invalid_config_exit_code():
+    """
+    If the config file contains unknown options we want to quit with a non-zero
+    exit-code
```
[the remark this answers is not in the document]
exhuma: Done in b2e76e6

tests/test_config.py
```diff
@@ -0,0 +1,180 @@
...
+
+    stderr = StringIO()
+    stdout = StringIO()
+    with patch("vulture.config.sys") as sys:
+        sys.stderr = stderr
+        sys.stdout = stdout
+        from_dict({"unknown_key_1": 1, "unknown_key_2": 1})
+    assert stdout.getvalue() == ""
+    assert "unknown_key_1" in stderr.getvalue()
+    assert "unknown_key_2" in stderr.getvalue()
```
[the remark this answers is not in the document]
exhuma: Done in 3ddc975

tox.ini
```diff
@@ -14,6 +14,7 @@ deps =
   coverage
   pytest
   pytest-cov
+  mock; python_version < "3.0"
```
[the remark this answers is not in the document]
exhuma: Done in bb9f8f2

vulture/config.py
```diff
@@ -0,0 +1,194 @@
...
+    # real default" in the CLI args, they should not be taken into
+    # consideration.
+    cli_config = _parse_args(argv)
+    for key in config.keys():
+        cli_value = cli_config.get(key, NO_DEFAULT)
+        if cli_value is not NO_DEFAULT:
+            config[key] = cli_value
+
+    if detected_toml_path and config["verbose"]:
+        print("Reading config values from {}".format(detected_toml_path))
```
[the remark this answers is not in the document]
exhuma: Done in 520f487

vulture/config.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def from_dict(data):
+    """
+    Create a new config dictionary from an existing one, assign possible
+    defaults and warn about unprocessed options.
+    """
+    # keep a copy of the keys, so we can keep track of any unprocessed
+    # values.
+    remaining_keys = set(data.keys())
```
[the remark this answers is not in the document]
exhuma: Done in a8a0138

For this I also needed to set default values on `argparse` arguments which did not yet have one or it would have given an error with the `NoneType`.

vulture/config.py
exhuma: sys.exit does not seem to write onto `stderr` (at least the test with `capsys` fails). So I kept the prints in there. But otherwise I implemented your suggestion in 3e21554

exhuma:
With the latest push I also merged with upstream to take advantage of the Python 3 stuff and cleaned out the remainders from tox which I added in earlier commits.

.github/workflows/main.yml
```diff
@@ -29,6 +29,10 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install coveralls setuptools tox wheel
 
+    - name: Install mock (For Python < 3.3)
+      if: ${{ matrix.python_version == '2.7' }}
+      run: "python -m pip install mock"
```
jendrikseipp, line 34: Thanks for making the changes! I'll have a more detailed look later, but it seems that this mock code can go away and that there's still a merge conflict.

exhuma: Ah right. Forgot about that. I currently don't get any merge conflicts. After merging your `master` into my feature branch yesterday I fixed all the conflicts and now I don't see anything when merging `upstream/master` into my `master`. Might still be something in the other direction though (even though GitHub does not show any conflicts either :thinking:).

I will try to rebase later today. If it's not getting too gnarly I can force-push after that giving a cleaner history. In the same go I can drop the py2/3 fixes as long as my commits were atomic enough :)

What's your thought on the rebase?

jendrikseipp: My bad, I was just confused by the GitHub UI. No need to rebase.

jendrikseipp: Before merging, I'll squash your changes into one commit, so you don't need to clean up the history.

vulture/config.py
```diff
@@ -0,0 +1,214 @@
...
+
+def from_dict(data):
+    """
+    Create a new config dictionary from an existing one, assign possible
+    defaults and warn about unprocessed options.
+    """
+    _check_config_types(data)
+
+    # keep a copy of the keys, so we can keep track of any unprocessed
+    # values.
```
jendrikseipp, line 55: The comment can go away.

vulture/config.py
```diff
@@ -0,0 +1,214 @@
...
+        (key, type(value))
+        for key, value in data.items()
+        if key in DEFAULTS and value is not NO_DEFAULT
+    }
+    expected_types = {(key, type(value)) for key, value in DEFAULTS.items()}
+    incorrect_types = given_types - expected_types
+    if incorrect_types:
+        keys = ", ".join(sorted(item[0] for item in incorrect_types))
+        msg = f"Incorrect data-types for the config values: {keys}"
+        sys.exit(msg)
```
jendrikseipp, line 44: Let's rename this function to _check_input_config and combine the key and type checking: 

``` python
for key, value in data.items():
    if key not in DEFAULTS:
        sys.exit(f"Unknown configuration key: {key}")
    if value is not NO_DEFAULT and type(value) != type(DEFAULTS[key]):
        sys.exit(f"Data type for {key} must be {DEFAULTS[key]}")
```

vulture/config.py
```diff
@@ -0,0 +1,214 @@
...
+        default=[],
+        help="Comma-separated list of paths to ignore (e.g.,"
+        ' "*settings.py,docs/*.py"). {glob_help} A PATTERN without glob'
+        " wildcards is treated as *PATTERN*.".format(**locals()),
+    )
+    parser.add_argument(
+        "--ignore-decorators",
+        metavar="PATTERNS",
+        type=csv,
+        default=[],
```
jendrikseipp, line 133: It would be good if we accessed DEFAULTS for retrieving the defaults everywhere.

vulture/config.py
exhuma: Done in 0fce022

vulture/config.py
exhuma: Done in 56c4f66

I still kept the print to stderr because I prefer those kinds of messages on stderr in case stdout is redirected. May happen in CI pipelines.

vulture/config.py
exhuma: Good..... point.... :smile: 

Sone in 3ba6473

tests/test_config.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    If the config file contains unknown options we want to see them on stderr.
+    We also should not see anything on stdout
+    """
+
+    with pytest.raises(SystemExit):
+        from_dict({"unknown_key_1": 1})
+    stdout, stderr = capsys.readouterr()
+    assert stdout == ""
+    assert "unknown_key_1" in stderr
```
jendrikseipp, line 177: Can't we combine the two tests above into one test?

.github/workflows/main.yml
```diff
@@ -29,6 +29,10 @@ jobs:
...
+
```
jendrikseipp, line 35: The mock-related code can go away.

vulture/config.py
jendrikseipp: Sounds good to me.

exhuma:
> Thanks for bearing with me :-)

No worries. It's amusing to be on the other side of pull requests for once. And seeing that there are people who are *even pickier than me* gives me some chuckles. Usually I'm the one with that kind of feedback, so I understand where you're coming from :smile:

tests/test_config.py
```diff
@@ -0,0 +1,191 @@
...
+)
+
+#: This dictionary maps types to "incompatible" types in the config. It is used
+#: later for parametrizing a test.
+WRONG_TYPES = {
+    int: str,
+    str: int,
+    list: str,
+    bool: str,
+}
```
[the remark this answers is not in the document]
exhuma: Initially I uses `type(x) is type(y)` for the type-checking but flake8 complained about that even though it is arguably more correct than an `isinstance` check. I did not want to add a `noqa` comment as you mentioned earlier you dislike them.

However, using `itertools` introduces a type-mapping between `int` and `bool` where the `isinstance` check will allow something that it shouldn't.

I have now added the `itertools` permutations in 03d0887 but this forced me to add a `noqa` comment. What's your thought on this?

tests/test_config.py
```diff
@@ -0,0 +1,191 @@
...
+        [tool.vulture]
+        verbose = true
+        """
+        )
+    )
+    cliargs = [
+        "cli_path",
+    ]
+    result = make_config(cliargs, toml)
+    assert result["verbose"] is True
```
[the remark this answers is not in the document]
exhuma: I do think this is already the case. If you look at https://github.com/exhuma/vulture/blob/feature/164-config-file-support/vulture/config.py#L187 it seems to me that it follows the ordering you mention.

Also, the unit-test just above the one you linked already tests all values. See https://github.com/exhuma/vulture/blob/feature/164-config-file-support/tests/test_config.py#L87

Let me know if I missed something.

vulture/config.py
```diff
@@ -0,0 +1,212 @@
...
+    def csv(exclude):
+        return exclude.split(",")
+
+    usage = "%(prog)s [options] PATH [PATH ...]"
+    version = f"vulture {__version__}"
+    glob_help = "Patterns may contain glob wildcards (*, ?, [abc], [!abc])."
+    parser = argparse.ArgumentParser(prog="vulture", usage=usage)
+    parser.add_argument(
+        "paths",
+        nargs="+",
```
jendrikseipp, line 113: The "+" requires users to specify paths on the command line. They can't do it in the TOML file. I think we should remove this requirement and update the `usage` string above on the way. Also, we should check that after reading all config values, there is at least one path and exit with a useful error message otherwise (if that isn't done already). Finally, this argument should also get `default=MISSING`.

vulture/config.py
```diff
@@ -0,0 +1,212 @@
...
+    """
+    _check_input_config(data)
+
+    unknown_keys = set(data) - set(DEFAULTS)
+    if unknown_keys:
+        print(
+            f"Unknown configuration keys: {sorted(unknown_keys)}",
+            file=sys.stderr,
+        )
+        sys.exit("invalid Config")
```
jendrikseipp, line 60: According to the Python docs sys.exit prints to stderr, so it seems the three calls above can be simplified: https://docs.python.org/3/library/sys.html#sys.exit

vulture/config.py
```diff
@@ -0,0 +1,212 @@
...
+            detected_toml_path = toml_path
+        else:
+            config = {}
+
+    # We can't use a simple call to dict.update() because some values should
+    # not be overwritten in the config. More precisely, if the values have "no
+    # real default" in the CLI args, they should not be taken into
+    # consideration.
+    cli_config = _parse_args(argv)
+    for key in config.keys():
```
jendrikseipp, line 204: I think it's clearer to use `for key, value in cli_config.items():` here.

tests/test_config.py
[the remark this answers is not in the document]
jendrikseipp: You're right, I don't see a logic error anymore.

tests/test_config.py
```diff
@@ -0,0 +1,178 @@
...
+    DEFAULTS,
+    _parse_args,
+    _parse_toml,
+    from_dict,
+    make_config,
+)
+
+#: This dictionary maps types to "incompatible" types in the config. It is used
+#: later for parametrizing a test.
+WRONG_TYPES = dict(permutations([int, str, list, bool], 2))
```
jendrikseipp, line 19: This only yields `{<class 'int'>: <class 'bool'>, <class 'str'>: <class 'bool'>, <class 'list'>: <class 'bool'>, <class 'bool'>: <class 'list'>}` and not all pairwise combinations of types.

vulture/config.py
```diff
@@ -0,0 +1,212 @@
...
+def _check_input_config(data):
+    """
+    Checks the types of the values in *data* against the expected types of
+    config-values. If a value is of the wrong type it will raise a SystemExit.
+    """
+    for key, value in data.items():
+        if key not in DEFAULTS:
+            print(f"Unknown configuration key: {key}", file=sys.stderr)
+            sys.exit("Invalid config")
+        if value is not MISSING and not type(value) is type(  # noqa: E721
```
jendrikseipp, line 38: I think it's a pity that there's no better way to solve this, but it's ok to leave the noqa there. Although I'd be interested to see an example of where isinstance screws things up here.

exhuma: The issue is with `int` and `bool`. Because `isinstance(False, int)` returns `True`.

jendrikseipp: Ah, thanks!

tests/test_config.py
exhuma: I'm currently working on this and personally don't like the complexity this introduces into the unit-tests. Especially the "parametrize" logic. I can simplify it and move the complexity from the `parametrize` setup *into* the unit-test but would break the "one check per test" rule in a way.

I feel 50/50 about going into so much details testing this. It's not like this is critical user-input that could generate code-injections or the like, and if a value has the wrong type it will crash out anyway on run.

I will postpone this for now and look at the other issues, let me know your decision on these tests.

vulture/config.py
exhuma: That won't work, because I don't *read* `value` from the dict. Instead I am using `key` to assign the CLI-args value into the dict.

tests/test_config.py
jendrikseipp: I agree that I steered you into the wrong direction there, sorry. Let's keep things simple as you originally proposed and just test some manual combinations of bug-prone type pairs.

vulture/config.py
jendrikseipp: I still think that the following code is identical and a bit simpler:

```
for key, value in cli_config.items():
    if value is not MISSING:
        config[key] = value
```

Am I missing something?

vulture/config.py
```diff
@@ -0,0 +1,212 @@
...
+    "exclude": [],
+    "ignore_decorators": [],
+    "ignore_names": [],
+    "make_whitelist": False,
+    "sort_by_size": False,
+    "verbose": False,
+}
+
+#: sentinel value to distinguish between "False" and "no default given"
+MISSING = object()
```
jendrikseipp, line 26: Instead of having this global variable, we could just delete all keys with MISSING values from the CLI dict before returning it in _parse_args(). If you don't want to do this change, I can also do it after merging.

vulture/config.py
exhuma: I had to resort to `prints` as using `sys.stderr` breaks the unit-tests with the `capsys` fixture. When using `sys.exit(str)`, both `stdout` and `stderr` values from `capsys` are empty.

I think it's a valid test to ensure we see something helpful on the console output in this test and would prefer to keep the prints. But it's up to you, if you want me to remove them, I can, and it will still print something to `stderr`, but it won't be testable.

jendrikseipp: I don't think we need to test the output.

exhuma: I now removed the print lines and am now confronted with a failing unit-test which fails and I don't know why. As the print statements are gone, it seems that even pytest is unable to see any output now.

A *personal* opinion I have held back until now is that I generally really dislike seeing `sys.exit` lines. I strongly prefer return values and just one `sys.exit` in the application entry-point. `sys.exit` often kneecaps instrumentation tools and makes debugging harder.

Additionally, I prefer to uses error-codes from [`errno.h`](https://python.readthedocs.io/en/latest/library/errno.html). But this is something I see so rarely done that I'm not hung up on this. Considering that it is used so rarely, it's not a very reliable listing anyway. Even Python itself does not return `ENOENT` on a `FileNotFoundError`.

In any case, removing `sys.exit` at this point would imply too many code-changes. If it is removed you suddenly have to think about return-values which will certainly break many things. And those return values (including exceptions) would bubble up and need to be handled somewhere appropriately.

I don't think this change is worth the effort (at least not in this PR). I will keep digging through the code and try to find out why the test fails.

vulture/config.py
exhuma: You're right. I misread your initial code snippet (mistook `cli_config.items` for `config.items()`

jendrikseipp:
Thanks a lot for all your work, @exhuma!

exhuma:
Thanks as well for bearing with me and for the merge. I just now wanted to remove the `MISSING` value when I saw you already merged it. It seemed trivial at first but the "point-in-time" when the defaults are assigned in (in `from_dict`) makes it harder than it seems to remove the `MISSING` value. There may yet be a simple solution, but I did not see it at first glance.

Also, sorry for my rant from yesterday. I had a bad day...

jendrikseipp:
No worries. This was a difficult feature to add, so I'm glad we managed to get it in.

PR #219, opened 2020-08-08
Treat `getattr(obj, "constant_string", default)` as a reference to `obj.constant_string`

jingw:
## Description
I realize handling `getattr` is a slippery slope. I think this case makes sense and is a reasonable place to draw the line.

My rationale is that the following code does the same thing, counts as a reference to obj.constant_string, but is not what most people would write.
```
try:
    x = obj.constant_string
except AttributeError:
    x = None
```

## Related Issue
Haven't filed one. Happy to discuss this makes sense or not.

## Checklist:
- [x] I have updated the documentation in the README.md file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Consider all files under `test` or `tests` directories test files
   (Jendrik Seipp).
 * Ignore `logging.Logger.propagate` attribute (Jendrik Seipp).
+* Treat `getattr/hasattr(obj, "constant_string", ...)` as a reference to
+  `obj.constant_string`. (jingw, #219)
 
 # 1.6 (2020-07-28)
 
--- a/tests/test_scavenging.py
+++ b/tests/test_scavenging.py
@@ -146,6 +146,30 @@ A._d_ = 4
     check(v.unused_vars, [])
 
 
+def test_getattr(v):
+    v.scan(
+        """\
+class Thing:
+    used_attr1 = 1
+    used_attr2 = 2
+    used_attr3 = 3
+    unused_attr = 4
+
+getattr(Thing, "used_attr1")
+getattr(Thing, "used_attr2", None)
+hasattr(Thing, "used_attr3")
+
+# Weird calls ignored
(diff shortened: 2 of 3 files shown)


vulture/core.py
```diff
@@ -508,6 +508,17 @@ def visit_Attribute(self, node):
         elif isinstance(node.ctx, ast.Load):
             self.used_attrs.add(node.attr)
 
+    def visit_Call(self, node) -> None:
+        """Count getattr(x, "some_attr", None) as a usage of some_attr"""
+        if (
+            isinstance(node.func, ast.Name)
+            and node.func.id == "getattr"
```
jendrikseipp, line 515: I think it would make sense to handle `hasattr` in the same way.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # unreleased
 
 * Bump minimum Python version to 3.6 (Jendrik Seipp, #218).
+* Treat `getattr(obj, "constant_string", default)` as a reference to `obj.constant_string`.
```
jendrikseipp, line 4: Feel free to add your name in brackets if you like.

vulture/core.py
```diff
@@ -508,6 +508,17 @@ def visit_Attribute(self, node):
         elif isinstance(node.ctx, ast.Load):
             self.used_attrs.add(node.attr)
 
+    def visit_Call(self, node) -> None:
```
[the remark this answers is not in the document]
jingw: whoops, left this type annotation by accident. removing.

tests/test_scavenging.py
```diff
@@ -146,6 +146,30 @@ def test_ignored_attributes(v):
...
 
 
+def test_getattr(v):
+    v.scan(
+        """\
+class Thing:
+    used_attr1 = 1
+    used_attr2 = 2
+    used_attr3 = 2
+    unused_attr = 3
```
jendrikseipp, line 156: Let's make the values go from 1 to 4 (which I guess you wanted to have anyway).

jendrikseipp:
Thanks for the patch!

PR #220, opened 2020-08-08
Parse mypy type comments

### LINKED ISSUE #152: Ignore type checking imports

jingw:
## Description

This prevents type-only imports from being flagged as unused.

Implementation decision note: As typed_ast will not be updated to support parsing Python 3.8+ syntax, I'm choosing not to use it. The downside of using Python's built-in type comment support is that it requires Python 3.8+. Someone more motivated than me could potentially make vulture use both, but I don't feel it's worth the complexity/effort given typed_ast seems likely to eventually die.

## Related Issue
Fixes #152
See also #153

## Checklist:
- [x] I have updated the documentation in the README.md file. (but do you want this in the README? not sure how much you want there versus left to --help)

### DIFF
--- a/tests/test_scavenging.py
+++ b/tests/test_scavenging.py
@@ -1,3 +1,5 @@
+import sys
+
 from . import check, v
 
 assert v  # Silence pyflakes.
@@ -644,16 +646,46 @@ x: List[int] = [1]
 def test_type_hint_comments(v):
     v.scan(
         """\
-import typing
+from typing import Any, Dict, List, Text, Tuple
 
-if typing.TYPE_CHECKING:
-    from typing import List, Text
 
-def foo(foo_li):
-    # type: (List[Text]) -> None
-    for bar in foo_li:
-        bar.xyz()
+def plain_function(arg):
+    # type: (Text) -> None
+    pass
+
+async def async_function(arg):
+    # type: (List[int]) -> None
+    pass
+
+some_var = {}  # type: Dict[str, str]
+
+class Thing:
+    def __init__(self):
+        self.some_attr = (1, 2)  # type: Tuple[int, int]
+
+for x in []:  # type: Any
+    print(x)
 """
     )
 
-    check(v.unused_imports, ["List", "Text"])
+    if sys.version_info < (3, 8):
+        check(v.unused_imports, ["Any", "Dict", "List", "Text", "Tuple"])
+    else:
+        check(v.unused_imports, [])
+        assert not v.found_dead_code_or_error
+
+
+def test_invalid_type_comment(v):
+    v.scan(
+        """\
+def bad():
+    # type: bogus
+    pass
+bad()
+"""
+    )
+
+    if sys.version_info < (3, 8):
+        assert not v.found_dead_code_or_error
+    else:
+        assert v.found_dead_code_or_error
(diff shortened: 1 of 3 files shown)


jingw:
Now that I look around, I see flake8 complains about bad type comments by default, without any obvious option to disable that. I'll go ahead and make the proposed change.

Originally, I decided to mirror the logic in https://bugs.python.org/issue35766 (the PR that added type_comments to ast.parse)
> By default, `ast.parse()` should not return type comments, since this would reject some perfectly good Python code (with sonething looking like a type comment in a place where the grammar doesn’t allow it). But passing an new flag will cause the tokenizer to process type comments and the returned tree will contain them.

That said, I imagine vulture does not need to be as careful as the standard library with regards to backwards compatibility.

jingw:
Done! I've made it automatically gate on Python 3.8+

vulture/core.py
```diff
@@ -218,7 +217,19 @@ def scan(self, code, filename=""):
...
             self.found_dead_code_or_error = True
         else:
-            self.visit(node)
+            # When parsing type comments, visiting can throw SyntaxError.
+            try:
+                self.visit(node)
+            except SyntaxError as err:
+                self._handle_syntax_error(err, filename)
+
+    def _handle_syntax_error(self, err, filename):
```
jendrikseipp, line 226: Let's make this a local function in the scan() method to encapsulate it better. Note that then the filename doesn't need to be passed.

vulture/core.py
```diff
@@ -601,6 +612,18 @@ def visit(self, node):
...
+        # There isn't a clean subset of node types that might have type
+        # comments, so just check all of them.
+        type_comment = getattr(node, "type_comment", None)
+        if type_comment is not None:
+            mode = (
+                "func_type"
+                if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef))
+                else "eval"
+            )
+            self.visit(ast.parse(type_comment, "<type_comment>", mode))
```
jendrikseipp, line 625: Let's use keys to make the call clearer: `ast.parse(type_comment, filename="<type_comment>", mode=mode)`

tests/test_scavenging.py
```diff
@@ -644,16 +646,46 @@ def test_var_type_annotation(v):
...
+    pass
+bad()
 """
     )
 
-    check(v.unused_imports, ["List", "Text"])
+    if sys.version_info < (3, 8):
+        assert not v.found_dead_code_or_error
+    else:
+        assert v.found_dead_code_or_error
```
jendrikseipp, line 691: Let's use `assert v.found_dead_code_or_error == sys.version_info >= (3, 8)`.

jingw: I actually find that version a bit harder to parse. It's a little bit trickier to translate to "on py3.8+, expect one behavior, else expect other behavior"

jendrikseipp: Point taken, let's stick with your version.

jendrikseipp:
Thanks a lot for your work on this!

djh82:
Apologies for being the PR necromancer; what was the rationale behind only enabling this for py3.8+?  Especially given 3.6+ supports typing (and was introduced in the now EOL 3.5)?  Thanks!

jendrikseipp:
Please see the issue description for the reasoning. In short: we don't think that the extra work is worth it.

Issue #221, opened 2020-08-08
False positive due to distinction between used_attrs and used_names


jingw:
Consider the following code in two files:
```
# module.py
some_module_var = 1
print(some_module_var)

# module_user.py
import module
module.some_module_var = 2
```

Running vulture complains that `module.some_module_var` is an unused attribute
```
$ vulture module.py module_user.py 
module_user.py:3: unused attribute 'some_module_var' (60% confidence)
```

How do you feel about this? On one hand, in practice, the cases where I've seen this false positive were test cases that should have been refactored to use `mock.patch` instead of directly setting module attributes. On the other hand, it's still a legitimate false positive, and someone could conceivably write non-test code like this.

I imagine other classes of false positives may be caused by separating attrs and names. One example that comes to mind is
```
class Foobar:
    def my_method(self):
        pass

    deprecated_name_for_my_method = my_method

Foobar().deprecated_name_for_my_method()
```

```
func.py:2: unused method 'my_method' (60% confidence)
```

Fixing this can be done by combining used_attrs with used_names in unused_attrs, though that largely defeats the point of separating these. I suppose it's not crazy to collapse them into a single used_names though.

jendrikseipp:
Good point! I agree that folding used_attrs into used_names makes sense (i.e., storing all used attributes in used_names and dropping the used_attrs set and the get_set function). Do you want to create a pull request for this?

jingw:
Yep, I can put together that change.

PR #223, opened 2020-08-14
Combine used_attrs and used_names to avoid false positives

jingw:
## Description
This fixes false positives when assigning to `x.some_name` but reading via `some_name`.
Note this produces false negatives in the opposite situation when `x.some_name` and `some_name` are different, so this is a trade-off in the direction of more precision / less sensitivity.

This direction is in line with pyflakes design principles: https://github.com/PyCQA/pyflakes#design-principles
> it will try very, very hard to never emit false positives.

Fixing this more accurately probably requires a lot more work to trace references and analyze context.

## Related Issue
#221

## Checklist:
- N/A I have updated the documentation in the README.md file.

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -177,9 +177,6 @@ class Vulture(ast.NodeVisitor):
         def get_list(typ):
             return utils.LoggingList(typ, self.verbose)
 
-        def get_set(typ):
-            return utils.LoggingSet(typ, self.verbose)
-
         self.defined_attrs = get_list("attribute")
         self.defined_classes = get_list("class")
         self.defined_funcs = get_list("function")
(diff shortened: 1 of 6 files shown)


vulture/core.py
```diff
@@ -461,7 +449,7 @@ def is_identifier(name):
                 self.used_names.add(name)
             for attr in name_and_attrs[1:]:
                 if is_identifier(attr):
-                    self.used_attrs.add(attr)
+                    self.used_names.add(attr)
```
jendrikseipp, line 452: The code for the name and the attributes can be combined now.

jingw: done!

jendrikseipp:
Thanks!

PR #225, opened 2020-08-26
Only parse format strings when being used with locals()

jingw:
## Description

The previous behavior has confusing results when strings aren't format strings, e.g. docstrings that happen to have a fragment resembling a format string. The new behavior is more precise.
    
The downside is that this won't detect usages when the format string or locals() is assigned to an intermediate variable before formatting. That seems fine to me, since it surprises me that the code `x = "{y}"` counts as usage of `y`. One could also argue against handling this at all, given that locals() is somewhat magical and that f-strings are the future.

## Related Issue
Haven't filed one. Happy to discuss.

## Checklist:
- (N/A) I have updated the documentation in the README.md file.

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -418,37 +418,6 @@ class Vulture(ast.NodeVisitor):
                     confidence=100,
                 )
 
-    def _handle_string(self, s):
-        """
-        Parse variable names in format strings:
-
-        '%(my_var)s' % locals()
-        '{my_var}'.format(**locals())
-        f'{my_var}'
-
-        """
-        # Old format strings.
-        self.used_names |= set(re.findall(r"\%\((\w+)\)", s))
-
-        def is_identifier(name):
-            return bool(re.match(r"[a-zA-Z_][a-zA-Z0-9_]*", name))
-
-        # New format strings.
-        parser = string.Formatter()
-        try:
-            names = [name for _, name, _, _ in parser.parse(s) if name]
-        except ValueError:
-            # Invalid format string.
-            names = []
-
-        for field_name in names:
-            # Remove brackets and their contents: "a[0][b].c[d].e" -> "a.c.e",
-            # then split the resulting string: "a.b.c" -> ["a", "b", "c"]
-            vars = re.sub(r"\[\w*\]", "", field_name).split(".")
-            for var in vars:
-                if is_identifier(var):
-                    self.used_names.add(var)
-
     def _define(
         self,
         collection,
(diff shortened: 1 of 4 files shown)


vulture/core.py
```diff
@@ -518,6 +487,58 @@ def visit_Call(self, node):
...
+        # '{my_var}'.format(**locals())
+        if (
+            isinstance(node.func, ast.Attribute)
+            and isinstance(node.func.value, ast.Str)
+            and node.func.attr == "format"
+        ):
+            if any(
+                kw.arg is None and self._is_locals_call(kw.value)
+                for kw in node.keywords
+            ):
```
jendrikseipp, line 500: Let's combine the two if-statements.

vulture/core.py
```diff
@@ -518,6 +487,58 @@ def visit_Call(self, node):
             if isinstance(attr_name_arg, ast.Str):
                 self.used_names.add(attr_name_arg.s)
 
+        # Parse variable names in new format strings:
+        # '{my_var}'.format(**locals())
```
jendrikseipp, line 491: Let's use double-quotes in example code as well.

vulture/core.py
```diff
@@ -518,6 +487,58 @@ def visit_Call(self, node):
...
+
+        for field_name in names:
+            # Remove brackets and their contents: "a[0][b].c[d].e" -> "a.c.e",
+            # then split the resulting string: "a.b.c" -> ["a", "b", "c"]
+            vars = re.sub(r"\[\w*\]", "", field_name).split(".")
+            for var in vars:
+                if is_identifier(var):
+                    self.used_names.add(var)
+
+    def visit_BinOp(self, node):
```
jendrikseipp, line 522: Please move this method above visit_Call to preserve alphabetical order.

vulture/core.py
```diff
@@ -518,6 +487,58 @@ def visit_Call(self, node):
...
+        """
+        Parse variable names in old format strings:
+
+        '%(my_var)s' % locals()
+        """
+        if isinstance(node.left, ast.Str) and self._is_locals_call(node.right):
```
jendrikseipp, line 528: Let's test that the BinOp operator is "%".

jendrikseipp (changes requested):
I like the change! Just some minor nitpicks as usual :-)

vulture/core.py
jingw: nice catch, updated the tests as well to cover this omission

jingw:
I'm not 100% sure what's up with the failing test. It's complaining that a file should be reformatted, but on a file I didn't touch.
```
style run-test: commands[0] | black --check --diff .
--- tests/test_config.py	2020-08-27 01:02:40.913556 +0000
+++ tests/test_config.py	2020-08-27 01:03:08.744364 +0000
@@ -155,11 +155,12 @@
     with pytest.raises(SystemExit):
         _check_input_config({"unknown_key_1": 1})
 
 
 @pytest.mark.parametrize(
-    "key, value", list(DEFAULTS.items()),
+    "key, value",
+    list(DEFAULTS.items()),
 )
 def test_incompatible_option_type(key, value):
     """
     If a config value has a different type from the default value we abort.
     """
would reformat tests/test_config.py
Oh no! 💥 💔 💥
1 file would be reformatted, 38 files would be left unchanged.
```

There appears to be something going on with black==20.8b1 versus 19.10b0. Version 20.8b1 was apparently just released. I don't see anything pinning a black version, but I also don't see why `tox` is giving me 19.10b0 after nuking the virtualenv and recreating it.

jingw:
Went ahead and appeased black's behavior change, stepping around "support for explicit trailing commas" :shrug:

jendrikseipp:
Thanks for the patch!

I'm not sure how to handle things like the black error in the future. Would you recommend pinning the version for linters like black and flake8 in the tox file?

jingw:
Both options are annoying IMO. The status quo makes CI break when something changes in a dependency, whereas pinning leaves you with stale versions unless you manually upgrade stuff periodically. (For personal projects, I don't bother pinning and just accept occasional breakages. For my job, breaking CI is disruptive, so everything's pinned.)

jendrikseipp:
I agree. I'll leave things as they are for now. Once dependabot supports tox.ini files, I'll probably try using it to create PRs for bumping versions automatically.

PR #224, opened 2020-08-20
Setup py

ju-sh:
## Description

I had a few simple suggestions regarding the setup.py file.
  
 - Use pathlib instead of os.path
 - Remove usage of codecs.open
 - Since `read()` is not used anywhere else, combine it with `find_version()`.

## Related Issue

Didn't open an issue for this since the changes were minor.                                                         

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have updated the documentation in the README.md file.
- [ ] I have added an entry in CHANGELOG.md.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #226, opened 2020-08-29
Use pathlib

ju-sh:
Use pathlib

## Description

 - Use pathlib for path operations.
 - ~~Add some type annotations.~~

## Related Issue

Mentioned at https://github.com/jendrikseipp/vulture/pull/224.

## Checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,11 +5,13 @@
 
 ## Related Issue
 <!--- Ideally, new features and changes are discussed in an issue first. -->
-<!--- If there is a corresponding issue, link to it here. Otherwise, remove this section. -->
+<!--- If there is a corresponding issue, please link to it here: -->
 
 ## Checklist:
-<!--- Go over the following points, and put an `x` into all boxes that apply. -->
-
-- [ ] I have updated the documentation in the README.md file or my changes don't require an update.
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->
+- [ ] My code follows the code style of this project.
+- [ ] I have updated the documentation in the README.md file.
 - [ ] I have added an entry in CHANGELOG.md.
-- [ ] I have added or adapted tests to cover my changes.
+- [ ] I have added tests to cover my changes.
+- [ ] All new and existing tests passed.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,13 +5,11 @@
 
 ## Related Issue
 <!--- Ideally, new features and changes are discussed in an issue first. -->
-<!--- If there is a corresponding issue, please link to it here: -->
+<!--- If there is a corresponding issue, link to it here. Otherwise, remove this section. -->
 
 ## Checklist:
-<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
-<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->
-- [ ] My code follows the code style of this project.
-- [ ] I have updated the documentation in the README.md file.
+<!--- Go over the following points, and put an `x` into all boxes that apply. -->
+
+- [ ] I have updated the documentation in the README.md file or my changes don't require an update.
 - [ ] I have added an entry in CHANGELOG.md.
-- [ ] I have added tests to cover my changes.
-- [ ] All new and existing tests passed.
+- [ ] I have added or adapted tests to cover my changes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# unreleased
+
+* Only parse format strings when being used with `locals()` (jingw, #225).
+* Don't override paths in pyproject.toml with empty CLI paths (bcbnz, #228).
+
 # 2.1 (2020-08-19)
 
 * Treat `getattr/hasattr(obj, "constant_string", ...)` as a reference to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * Only parse format strings when being used with `locals()` (jingw, #225).
 * Don't override paths in pyproject.toml with empty CLI paths (bcbnz, #228).
+* Use pathlib internally (ju-sh, #226).
 
 # 2.1 (2020-08-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,3 @@
-# unreleased
-
-* Only parse format strings when being used with `locals()` (jingw, #225).
-* Don't override paths in pyproject.toml with empty CLI paths (bcbnz, #228).
-
 # 2.1 (2020-08-19)
 
 * Treat `getattr/hasattr(obj, "constant_string", ...)` as a reference to
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -7,7 +7,7 @@ import pytest
 from vulture import core
 
 REPO = pathlib.Path(__file__).resolve().parents[1]
-WHITELISTS = (REPO / "vulture" / "whitelists").glob("*.py")
+WHITELISTS = list((REPO / "vulture" / "whitelists").glob("*.py"))
 
 
 def call_vulture(args, **kwargs):
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,5 +1,5 @@
 import glob
-import os.path
+import pathlib
 import subprocess
 import sys
 
@@ -7,9 +7,8 @@ import pytest
 
 from vulture import core
 
-DIR = os.path.dirname(os.path.abspath(__file__))
-REPO = os.path.dirname(DIR)
-WHITELISTS = glob.glob(os.path.join(REPO, "vulture", "whitelists", "*.py"))
+REPO = pathlib.Path(__file__).resolve().parents[1]
+WHITELISTS = glob.glob(str(REPO / "vulture" / "whitelists" / "*.py"))
 
 
 def call_vulture(args, **kwargs):
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -7,7 +7,9 @@ import pytest
 from vulture import core
 
 REPO = pathlib.Path(__file__).resolve().parents[1]
-WHITELISTS = list((REPO / "vulture" / "whitelists").glob("*.py"))
+WHITELISTS = [
+    str(path) for path in (REPO / "vulture" / "whitelists").glob("*.py")
+]
 
 
 def call_vulture(args, **kwargs):
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,4 +1,3 @@
-import glob
 import pathlib
 import subprocess
 import sys
@@ -8,7 +7,7 @@ import pytest
 from vulture import core
 
 REPO = pathlib.Path(__file__).resolve().parents[1]
-WHITELISTS = glob.glob(str(REPO / "vulture" / "whitelists" / "*.py"))
+WHITELISTS = (REPO / "vulture" / "whitelists").glob("*.py")
 
 
 def call_vulture(args, **kwargs):
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -147,6 +147,27 @@ def test_config_merging_missing():
     assert result["ignore_names"] == ["name1"]
 
 
+def test_config_merging_toml_paths_only():
+    """
+    If we have paths in the TOML but not on the CLI, the TOML paths should be
+    used.
+    """
+    toml = StringIO(
+        dedent(
+            """\
+        [tool.vulture]
+        paths = ["path1", "path2"]
+        """
+        )
+    )
+    cliargs = [
+        "--exclude=test_*.py",
+    ]
+    result = make_config(cliargs, toml)
+    assert result["paths"] == ["path1", "path2"]
+    assert result["exclude"] == ["test_*.py"]
+
+
 def test_invalid_config_options_output():
     """
     If the config file contains unknown options we want to abort.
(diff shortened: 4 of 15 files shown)


ju-sh:
Hi.

Coverage has reduced but I thought it was better to see if the changes were okay before adding to the tests as I have probably broken something.

tests/__init__.py
```diff
@@ -1,15 +1,14 @@
...
 
 import pytest
 
 from vulture import core
 
-DIR = os.path.dirname(os.path.abspath(__file__))
-REPO = os.path.dirname(DIR)
-WHITELISTS = glob.glob(os.path.join(REPO, "vulture", "whitelists", "*.py"))
+REPO = pathlib.Path(__file__).resolve().parents[1]
+WHITELISTS = glob.glob(str(REPO / "vulture" / "whitelists" / "*.py"))
```
jendrikseipp, line 11: I think it would be nice to use `Path.glob` here.

jendrikseipp (changes requested):
Thanks for the patch! I had a quick look and the pathlib code looks good. However, reviewing this in more detail is hard, because most of the changes relate to adding type annotations. I haven't made up my mind about whether type annotations are useful for Vulture completely (given that it's primarily used as an app, not a library and has 100% test coverage), but I'm leaning towards adding them, at least where they make understanding the code easier. Therefore, I'd ask you to remove the type annotations, type imports and changes only related to mypy for now. Maybe you can store the diff related to types somewhere, for a separate PR. Depending on how separated your commits are, this may be easiest to to with a graphical diff tool like "meld".

tests/__init__.py
ju-sh: Had missed that. Will change.

ju-sh:
Sorry for the delay. Got held up with something else in between.

I had ended up adding annotations to help me figure out the changes to make. I've removed them.

But, as an aside, I think it's good to have type annotations. They can really help a lot while making changes. Even when we don't do static type checking.

Let me know of further changes so we can go ahead and squash the commits.

tests/test_config.py
```diff
@@ -157,7 +157,8 @@ def test_invalid_config_options_output():
 
 
 @pytest.mark.parametrize(
-    "key, value", list(DEFAULTS.items()),
+    "key, value",
```
ju-sh, line 160: Hmm... I don't know what's going on here. I thought Black would've formatted it to looks similar? I mean, there are no changes here.

jendrikseipp: Don't worry about the black problems. I'll pin the black version to an earlier version and you can rebase from master if you don't mind. No hurry though, I won't have time to look at your changes before the weekend.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,13 +5,11 @@
 
 ## Related Issue
 <!--- Ideally, new features and changes are discussed in an issue first. -->
-<!--- If there is a corresponding issue, please link to it here: -->
+<!--- If there is a corresponding issue, link to it here. Otherwise, remove this section. -->
```
jendrikseipp, line 8: It would be nice if the pull request only showed the new changes. Can you rebase your changes please?

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# unreleased
+
+* Only parse format strings when being used with `locals()` (jingw, #225).
+* Don't override paths in pyproject.toml with empty CLI paths (bcbnz, #228).
```
jendrikseipp, line 4: After the rebase, please add a changelog entry, including your name and PR number if you like.

tests/test_encoding.py
```diff
@@ -26,22 +26,21 @@ def test_encoding2(v):
     assert not v.found_dead_code_or_error
 
 
-def test_non_utf8_encoding(v, tmpdir):
+def test_non_utf8_encoding(v, tmp_path):
```
jendrikseipp, line 29: I think the tmpdir name is needed as-is for pytest to work (it automatically creates a temporary directory).

jendrikseipp (changes requested):
After the rebase, please go over the diff and check that only pathlib-related changes are present. For example, I still saw some type annotations, which make reviewing hard.

tests/test_encoding.py
ju-sh: `tmp_path` does the same thing. It's like the pathlib compatible replacement for tmpdir. But got to figure out if it works different on windows.

.github/PULL_REQUEST_TEMPLATE.md
ju-sh: Hmm.. I figure I messed up the branch. I haven't done this before. How can I rebase like that? Can you guide me?

tests/test_encoding.py
jendrikseipp: Ah, thanks!

.github/PULL_REQUEST_TEMPLATE.md
jendrikseipp: I think the solution should be the following. origin should be the main vulture repo. 

```
git fetch origin
git checkout test_trial
git rebase origin/master
git push --force
```

(from https://stackoverflow.com/a/45286781/1277298 and https://github.community/t/github-pull-requests-showing-invalid-diff-for-already-merged-branch/559/3)

vulture/core.py
```diff
@@ -662,7 +686,7 @@ def generic_visit(self, node):
                 self.visit(value)
 
 
-def _parse_args():
+def _parse_args():  # XXX: two _parse_args() ???
```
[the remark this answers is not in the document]
jendrikseipp: Good catch! I removed the duplicate code.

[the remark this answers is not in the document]
ju-sh: I've removed the extra `__version__` variable as well.

CHANGELOG.md
jendrikseipp: If the rebase doesn't work, please start a new branch from master, copy your changes there and open a new pull request.

jendrikseipp: I think I fixed it now. The solution was to merge master into your branch. I'll have a look at the code now :-)

vulture/core.py
```diff
@@ -191,11 +200,11 @@ def get_list(typ):
         self.ignore_names = ignore_names or []
         self.ignore_decorators = ignore_decorators or []
 
-        self.filename = ""
+        self.filename = pathlib.Path()
         self.code = []
         self.found_dead_code_or_error = False
 
-    def scan(self, code, filename=""):
+    def scan(self, code, filename=DEFAULT_FILE_PATH):
```
jendrikseipp, line 207: filename can stay a string, because it's only used for printing and never for accessing files, I think.

vulture/utils.py
```diff
@@ -46,10 +47,11 @@ def condition_is_always_true(condition):
...
 
 def format_path(path):
-    if not path:
-        return path
-    relpath = os.path.relpath(path)
-    return relpath if not relpath.startswith("..") else path
+    path_str = str(path)
+    if not path_str:
+        return path_str
+    relpath = os.path.relpath(path_str)
```
jendrikseipp, line 53: Is there no pathlib alternative for this?

vulture/utils.py
```diff
@@ -63,23 +65,32 @@ def get_decorator_name(decorator):
     return "@" + ".".join(reversed(parts))
 
 
-def get_modules(paths, toplevel=True):
+def get_modules(paths):
```
jendrikseipp, line 68: How does the new code handle nested directories without recursion?

tests/test_script.py
```diff
@@ -7,7 +7,7 @@
 
 
 def test_module_with_explicit_whitelists():
-    assert call_vulture(["vulture/"] + WHITELISTS) == 0
+    assert call_vulture(["vulture/"] + list(WHITELISTS)) == 0
```
jendrikseipp, line 10: If WHITELISTS is a generator now, we should turn it into a list when it is created. Otherwise, we can drop the list() call here.

jendrikseipp (changes requested):
This looks much nicer :-) I think with my proposed changes, we can get rid of some pathlib->string conversions, which should make the code cleaner.

vulture/utils.py
ju-sh: There is relative_to but docs said they were not exactly equivalent. But it seems to be enough for our purpose. Has added it.

vulture/utils.py
ju-sh: The rglob() function does it.

vulture/core.py
```diff
@@ -232,14 +243,14 @@ def handle_syntax_error(e):
 
     def scavenge(self, paths, exclude=None):
```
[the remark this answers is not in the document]
ju-sh: You mean, convert `paths` to `Path` objects at the start of the function itself?

vulture/core.py
ju-sh: It would be used to find the relative path in `utils.format_path()` though..

vulture/core.py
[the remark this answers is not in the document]
jendrikseipp: Exactly. Usually, you want to convert to the internal representation (pathlib) as early as possible and convert back (to strings) as late as possible to avoid having to convert back and forth between the two representations.

vulture/core.py
jendrikseipp: Ah, good point! Then let's convert accept strings and Path objects for scan() as well and convert the argument to a Path object at the beginning of the method. The default `filename=""` (empty string) can remain unchanged, no need for DEFAULT_FILE_PATH.

vulture/utils.py
jendrikseipp: Ah, that's cool, didn't know about this function.

vulture/utils.py
```diff
@@ -46,10 +47,10 @@ def condition_is_always_true(condition):
 
 
 def format_path(path):
-    if not path:
+    try:
+        return path.relative_to(os.curdir)
+    except ValueError:
         return path
```
jendrikseipp, line 53: Let's add a short comment above `return path`: `# Path is not below the current directory.`

vulture/core.py
ju-sh: But with the changes in this PR, the `filename` arg only accepts `Path` objects. Should we still allow `str`s? Or maybe this got to be added for leaving provision for changes allowing `str` in the future?

Also, we could remove the `DEFAULT_FILE_PATH` and use `None` instead.

jendrikseipp: Yes, let's allow strings and Path objects. And remove DEFAULT_FILE_PATH and use `filename=""` as before.

vulture/core.py
```diff
@@ -42,13 +41,18 @@ def _is_special_name(name):
...
 def _match(name, patterns, case=True):
-    func = fnmatchcase if case else fnmatch
-    return any(func(name, pattern) for pattern in patterns)
+    if not case:
+        name = pathlib.Path(str(name).lower())
+    return any(name.match(pattern) for pattern in patterns)
+
+
+def _match_name(name, patterns):
+    return any(fnmatchcase(name, pattern) for pattern in patterns)
```
jendrikseipp, line 50: I think it's better to let _match() only accept strings (callers must then convert to strings themselves) and remove the _match_name() function again.

vulture/utils.py
```diff
@@ -63,23 +65,32 @@ def get_decorator_name(decorator):
...
+                if top_path.suffix == ".pyc":
+                    sys.exit(f".pyc files are not supported: {top_path}")
+                else:
+                    modules.append(top_path)
+            elif not top_path.is_dir():
+                sys.exit(f"Error: {top_path} could not be found.")
+        sub_paths = path.rglob("*.py")
+        for sub_path in sub_paths:
+            if sub_path.is_file():
+                modules.append(sub_path)
```
jendrikseipp, line 93: Could you please add some tests that check get_modules()?

vulture/utils.py
```diff
@@ -63,23 +65,32 @@ def get_decorator_name(decorator):
...
-        elif toplevel:
+        path = path.resolve()
+
+        if not path.exists():
             sys.exit(f"Error: {path} could not be found.")
+
+        if path.is_file():
+            top_paths = (_ for _ in [path])
+        else:
+            top_paths = path.glob("*")
```
jendrikseipp, line 80: The new code seems overly complicated. Basically, it should do: loop over all given paths, add all *files* that end with .py, abort if any ends with .pyc and add all .py files under the given *directories*.

tests/test_encoding.py
```diff
@@ -26,22 +26,21 @@ def test_encoding2(v):
...
     code = ""
     name = "non_utf8"
-    non_utf_8_file = str(tmpdir.mkdir(name).join(name + ".py"))
+    non_utf_8_file = tmp_path / (name + ".py")
     with open(non_utf_8_file, mode="wb") as f:
         f.write(codecs.BOM_UTF16_LE)
         f.write(code.encode("utf_16_le"))
-    v.scavenge([f.name])
+    v.scavenge([str(non_utf_8_file)])
```
jendrikseipp, line 36: Let's pass strings to scavenge() and scan() in the tests.

jendrikseipp (changes requested):

tests/test_encoding.py
jendrikseipp: This comment is confusing. What I mean is: in the tests, let's pass a Path object if we already have one (like in the code above) and let's pass strings in the other cases below (where we currently create Path objects just to comply with the API).

vulture/utils.py
ju-sh: The function did look long but I couldn't think of a simpler way.

What I tried to do there is as the way that you described.

Any ideas on making it simpler?

tests/test_encoding.py
ju-sh: Good point. Will do.

vulture/utils.py
jendrikseipp: I'll have a look when I find the time. Can you make sure that all tests pass?

vulture/core.py
```diff
@@ -48,8 +47,17 @@ def _match(name, patterns, case=True):
 
 def _is_test_file(filename):
     return _match(
-        os.path.abspath(filename),
-        ["*/test/*", "*/tests/*", "*/test*.py", "*/*_test.py", "*/*-test.py"],
+        filename.resolve(),
+        [
+            "*/test/*",
```
ju-sh, line 52: I added more patterns here to fix some windows errors. pathlib works differently in unix and windows it seems.

jendrikseipp:
Thanks for your work on this!

PR #228, opened 2020-09-08
Don't override TOML paths with empty CLI paths

bcbnz:
The paths CLI argument did not use the default=missing sentinel, and so
returned an empty list if no paths were given to the CLI. This meant any
paths configured in pyproject.toml were overridden with an empty list,
making the path= setting useless.

Adding the sentinel to the CLI fixes this behaviour. A regression test
is also added to make sure TOML paths are used if no CLI paths are
given.

## Checklist:

- [X] I have updated the documentation in the README.md file or my changes don't require an update.
- [X] I have added an entry in CHANGELOG.md.
- [X] I have added or adapted tests to cover my changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # unreleased
 
 * Only parse format strings when being used with `locals()` (jingw, #225).
+* Don't override paths in pyproject.toml with empty CLI paths (bcbnz, #228).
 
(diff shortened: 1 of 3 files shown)


bcbnz:
Force-pushed to fix style and add changelog entry with pull request ID.

jendrikseipp:
Thanks for reporting this and providing a clean fix!

Issue #227, opened 2020-09-06
option to output absolute pathnames of reported files


mcooperman:
Using vulture as an External Tool with PyCharm.
a linter, such as Flake8, outputs reported issues with absolute pathnames to the files. These are 'clickable' in the PyCharm Run UI, to jump directly to the related code.
vulture (after testing outside of pycharm as well) seems to only output relative paths from the cwd without a leading /
it appears that PyCharm relies on detecting paths in the Run output with the starting / (e.g. absolute path), for external tools anyway (Pylint has a more complex integration).
if vulture had a command line option to tell it to output absolute paths, that would be helpful.
I'd be happy to contribute.

jendrikseipp:
Thanks for the report! Ideally, I'd like to find a solution that works for multiple editors. Do you know which format would be required for other popular editors (https://insights.stackoverflow.com/survey/2019#development-environments-and-tools)?

Are you sure flake8 uses absolute paths? Or does it come with an option for this? For me flake8 3.8.1 shows relative paths.

PR #230, opened 2020-09-08
Issue #227, Feature/pathformat, adds --absolute-paths option

mcooperman:
## Description

by invoking with --absolute paths, output of problems found will specify file names in absolute path, platform-specific
when used with PyCharm, as an external tool, can now 'jump to code' on these messages by clicking on the file paths.

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ vulture.egg-info/
 .pytest_cache/
 .tox/
 .venv/
+.idea/
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -27,6 +27,7 @@ def test_cli_args():
         ignore_names=["name1", "name2"],
         make_whitelist=True,
         min_confidence=10,
+        absolute_paths=True,
         sort_by_size=True,
         verbose=True,
     )
@@ -39,6 +40,7 @@ def test_cli_args():
             "--min-confidence=10",
             "--sort-by-size",
             "--verbose",
+            "--absolute-paths",
             "path1",
             "path2",
         ]
@@ -97,6 +99,7 @@ def test_config_merging():
         min_confidence = 10
         sort_by_size = false
         verbose = false
+        absolute_paths = true
         paths = ["toml_path"]
         """
         )
@@ -108,6 +111,7 @@ def test_config_merging():
         "--make-whitelist",
         "--min-confidence=20",
         "--sort-by-size",
+        "--absolute-paths",
         "--verbose",
         "cli_path",
     ]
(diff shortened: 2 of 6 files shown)


mcooperman:
I have updated the documentation in the README.md file or my changes don't require an update.
 I have added an entry in CHANGELOG.md.
 I have added or adapted tests to cover my changes.

looked at flake8, which sports an extension, 'flake8_formatter_abspath' to do exactly this same thing, which suggests it is of general use...
I added this to tox.ini to make use of it for flake8 errors flagged in PyCharm when using flake8 as an external tool in the Run console of PyCharm

jendrikseipp:
Could you please state the steps that are necessary to use Vulture within PyCharm?

Instead of a binary flag, I think it would be better to use the same approach as flake8, i.e., a `--format` parameter that accepts "relative" and "absolute". Then it will be easier to add other formats in the future (e.g., file:// URIs for clickable links in GNOME Terminal or JSON output).

jendrikseipp (changes requested):
Please go over the PR and make sure to include only the minimal changes that are necessary for the new feature plus documentation and tests.

jendrikseipp (changes requested):
I'd like to wait with this until #226 is merged. 

I had a brief look at your patch and my first impression is that it's a little more complex than it needs to be. I don't think we need abstract base classes for example. Also, I prefer "--format" over "--path-format". Also, the patch shouldn't contain any "extras" like pylint comments.

mcooperman:
> I'd like to wait with this until #226 is merged.
> 
> I had a brief look at your patch and my first impression is that it's a little more complex than it needs to be. I don't think we need abstract base classes for example. Also, I prefer "--format" over "--path-format". Also, the patch shouldn't contain any "extras" like pylint comments.

sorry - missed those #pylints - i've become so used to them ;) 
will make the change to --format from --path-format
i'll make the format class base non-abstract, maybe default to the relative behavior
i did try to respond to the request to accommodate more flexibility in the future to specify other values for formats...

jendrikseipp:
Sounds good. I'll have a closer look once the other PR is merged and you merged the master branch into your branch.

jendrikseipp:
#226 is now merged. If you like, you can merge the master branch into your branch. No need to rebase, I'll squash all commits in the end anyway.

jendrikseipp (changes requested):
This pull request has become quite complicated. 376 new lines for such a simple feature is too much in my opinion. I think the same functionality could be obtained with a simpler design along the lines of this pseudo-code:

```
path = get path from item
if format == absolute:
    path.resolve()
print item with path to stdout
```

Can you slim down the pull request to this simpler code, please?

mcooperman:
didnt see your last comment. just pushed latest commits incl removing some stray debug prints and an attempted fix for windows errors on absolute path testing. once that's done will look at the simplification.

jendrikseipp:
Are you still planning to pursue this?

PR #234, opened 2020-10-02
Add clarification for the docs

### LINKED ISSUE #229: Clarify how config and CLI options are merged
As far as I can tell, if an option is given in both pyproject.toml and on the CLI, then the CLI version is used. This should be explicitly stated in the documentation, as it might be expected that options which take lists (e.g., `exclude`) would be concatenated.

exhuma:
Closes #229


jendrikseipp:
Thanks!

PR #236, opened 2020-10-07
Add tests for python3.9

ju-sh:
Add tests for python3.9

## Description
Test for the new Python version in tox and github actions.

The new version doesn't seem to break anything.

https://docs.python.org/3.9/whatsnew/3.9.html

## Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-]       python-version: [pypy3, 3.6, 3.7, 3.8, 3.9]
+        python-version: [pypy3, 3.6, 3.7, 3.8, 3.9]
         os: [macos-10.15, ubuntu-18.04, ubuntu-20.04, windows-2019]
 
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [pypy3, 3.6, 3.7, 3.8]
+]       python-version: [pypy3, 3.6, 3.7, 3.8, 3.9]
         os: [macos-10.15, ubuntu-18.04, ubuntu-20.04, windows-2019]
 
     steps:
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ setuptools.setup(
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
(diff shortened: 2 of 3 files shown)


jendrikseipp:
Nice, thanks!

jendrikseipp:
I was a bit quick to merge this... Due to a syntax error in the GitHub action, your tests never ran. After fixing the error, the test for Pypy on Windows fails (https://github.com/jendrikseipp/vulture/runs/1222113460?check_suite_focus=true). Could you look into this, please @ju-sh ? If you can't find a fix, please skip the Pypy on Windows test in the GitHub Action.

ju-sh:
Maybe making the test use a newer version of tox would fix the problem?

The tests seem to use tox 3.2.0 and the latest one is 3.13.2.

I had a look at tox's source and there is significant change in the file that gives error in the test. A windows directory itself has been added (as in [here](https://github.com/tox-dev/tox/tree/master/src/tox/interpreters)).

Link to the 3.2.0 version: https://github.com/tox-dev/tox/blob/6cc196301f054c8babff0df0bac4253c576555b9/src/tox/interpreters.py

jendrikseipp:
It seems this isn't related to your changes at all. Would yo care to investigate nonetheless?

ju-sh:
I looked through some pages, but couldn't find a fix for this. 

Maybe we moved too quick and perhaps some actions has incompatibility with new version maybe? Still the failing test had passed earlier with what seems to be the same pypy version.

I wonder what's going wrong.. :-(

ju-sh:
> It seems this isn't related to your changes at all. Would yo care to investigate nonetheless?

I'll try. Which github action error had you set right earlier though? Was it the setup python?

jendrikseipp:
I'm not sure what you mean. The error has nothing to do with our changes. Only with some version of a dependency which was updated behind our back. Could you try to pin the tox version to an earlier release?

ju-sh:
No you mentioned a syntax error. I just wanted to know where the error was in. Was just curious to know if it was in a github action's code or in the main.yml file.

jendrikseipp:
There was an extra [ in the main.yml file. I looked at the logs of the last
successful and the first failing build and the only difference is the minor
Windows version. I think the best strategy is to skip the tests for Pypy on
Windows for now. Can you try that, please?

ju-sh:
Shall I create a new PR after removing pypy on windows? Commits to the branch of this PR is not showing up here.

jendrikseipp:
Yes and no 😊

The-Compiler:
> And is there a way to try out github action without having to push to github every time?

FWIW there's [act](https://github.com/nektos/act), but I've never actually tried it, and it doesn't seem to support Windows.

jendrikseipp:
Thanks for the hint! Didn't know about this tool. It looks quite useful!

PR #237, opened 2020-10-12
skip test for pypy3 on windows

ju-sh:
Skip tests for pypy3 on windows

## Description
Skip tests for pypy3 on windows in tox and github actions as they seem to fail for reasons unrelated to the repo's code.

## Related

https://github.com/jendrikseipp/vulture/pull/236

## Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ jobs:
         os: [macos-10.15, ubuntu-18.04, ubuntu-20.04, windows-2019]
 
     steps:
-    - name: Checkout Code
+    - name: Checkout code
       uses: actions/checkout@v2
 
     - name: Set up Python ${{ matrix.python-version }}
@@ -24,7 +24,7 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install Dependencies
+    - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
         python -m pip install --upgrade coveralls setuptools tox wheel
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = cleanup, py{36,37,38}, style
+envlist = cleanup, py{36,37,38,39,py3}, style
```
jendrikseipp, line 2: The py3 should be unnecessary, right? Let's remove it.

ju-sh: Wouldn't that be needed for running tox locally with pypy? Or would that be included automatically?

jendrikseipp: No need to test pypy3 locally by default. We can still use "tox -e pypy3" locally if we want to test it.

ju-sh: Oh, okay. I'll remove that then.

.github/workflows/main.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         python -m vulture vulture/ tests/
 
     - name: Run Tests
+      if: ${{ matrix.python-version != 'pypy3' || matrix.os != 'windows-2019' }}
       run: python -m tox -e py
```
jendrikseipp, line 50: Maybe we can simply fix the test instead of skipping it. Can you try to use `tox -e pypy3` for Windows with PyPy and `tox -e py` for all other matrix combinations?

ju-sh: That didn't work either..

jendrikseipp: Thanks for testing this. I'll remove the test again and then I'll merge this.

jendrikseipp:
Thanks @ju-sh !

PR #239, opened 2020-10-27
Print a success message to stdout

exhuma:
## Description

I usually run vulture in an auto re-running sub-shell, and because vulture neither prints a startup, nor a status-message on success I can never tell whether vulture has actually run or not.

This change adds a simple status message on successful run. I did not add a failure message because this is already covered by the normal printout from vulture.

At first I simply wrote ``done``, but made it a bit more verbose to make it more clear that the message comes from vulture. I was alternatively thinking of making it ``[vulture] done`` but I think the "normal/natural" English I used makes it more consistent with the exiting messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Don't override paths in pyproject.toml with empty CLI paths (bcbnz, #228).
 * Run continuous integration tests for Python 3.9 (ju-sh, #232).
 * Use pathlib internally (ju-sh, #226).
+* Print a message to stdout on success (exhuma).
 
 # 2.1 (2020-08-19)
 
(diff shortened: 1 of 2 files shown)


jendrikseipp:
I appreciate the effort, but being silent when there is nothing to report is part of the Unix philosophy (http://www.linfo.org/rule_of_silence.html). Also, related tools such as flake8 or pycodestyle also print nothing when there are no warnings.

RJ722:
@exhuma If you still want the same output, you might can alias vulture with a shell script which looks a bit like this:

```sh
output=`vulture "$@"`

if [[ "$output" == false ]]; then
	echo "Successfully finished. No unused code detected."
else
	echo "$output"
fi
```

PR #245, opened 2021-01-16
doc: hotfix pre-commit instructions

neutrinoceros:
This is a hotfix to #244, fixing an important mistake in README.md


jendrikseipp:
Good catch, thanks!

pquentin:
Thanks, I was about to report that! (Note that the quotes aren't needed: since there's a `v` in the string, it cannot be mistaken for a number.)

neutrinoceros:
Thank you for the clarification, I never quite understood why it was sometimes necessary but not always :)

PR #247, opened 2021-01-19
Improve pytest invocation in tox

The-Compiler:
This avoids the old py.test name (replaced by pytest in v3.0.0 in August
2016), and also makes it possible to pass arguments to pytest like so:

    tox -e py39 -- tests/test_utils.py

(checklist not really applicable)


jendrikseipp:
Interesting! Didn't know about that feature, thanks!

PR #248, opened 2021-01-19
Add black to PR template checklist

The-Compiler:
Would've avoided one force push I had to do :slightly_smiling_face: 

(checklist not really applicable)


jendrikseipp:
Good idea, thanks!

Issue #208, opened 2020-05-28
Add support for flask-restful api functions


HitLuca:
As for right now, vulture detects as unused function my flask-restful functions (delete in this case). Would it be feasible to add the knowledge that flask-restful behaves in this way?

jendrikseipp:
Could you provide an example, please?

HitLuca:
file `test.py`

```python
from flask_restful import Resource


class Test(Resource):
    @staticmethod
    def delete():
        pass
```

`vulture test.py`

`test.py:5: unused function 'delete' (60% confidence)`

onlinejudge95:
+1 for API methods defined in following packages

- [Flask-RESTful](https://flask-restful.readthedocs.io/en/latest/)
- [Flask-RESTPlus](https://flask-restplus.readthedocs.io/en/stable/)
- [Flask-RESTX](https://flask-restx.readthedocs.io/en/latest/)

jendrikseipp:
This sounds like a good fit for a whitelist. If you like, we can try to integrate your whitelist for flask or flask-restful into Vulture. See the whitelists directory for examples.

HitLuca:
@jendrikseipp how would you go with adding an external package into your whitelist without having to import it? That would define an external dependency which I don't think you would like to have

RJ722:
@HitLuca because Vulture isn't aware of the package, or it's namespace, we don't need to actually source the actual module -- we just need a variable with the same name.

jendrikseipp:
Actually, i think it's good to use "real" source code in whitelists. That way, we can test that the names really exist and are spelled correctly. The whitelist files that are bundled with Vulture are only used when the corresponding module is imported, and even then the content is only parsed for names, not actually run, so no new dependency is needed. We only need to add flask or its extensions to the tox.ini testenv section and this allows us to actually run the whitelist file during testing.

HitLuca:
Makes sense, thank you. I'll write something down for flask when I have the time

PR #258, opened 2021-05-03
Add whitelist for the pint package

bje-:
Self-evident.


RJ722 (commented):
Thanks for the pull request @bje-! 

I think that we need to think some more about shipping whitelists for non-stdlib modules with Vulture. I can think of the following options:

1. We wrap the entire script in a try/except block which will only run if the module can be imported successfully.
2. Use the proxy `Whitelist` module from `whitelists/whitelist_utils.py`.

@jendrikseipp any thoughts?

---

I also had an initial worry that using common names (like in this case, `default_format`) might be used too often altogether and that Vulture might miss them as false negatives, but now that I think of it, since we only import the whitelists only if the corresponding module is imported in user's program, I don't think this will be a problem.

RJ722:
Oh I just found out https://github.com/jendrikseipp/vulture/issues/208#issuecomment-637369208 from @jendrikseipp:

> Actually, i think it's good to use "real" source code in whitelists. That way, we can test that the names really exist and are spelled correctly. The whitelist files that are bundled with Vulture are only used when the corresponding module is imported, and even then the content is only parsed for names, not actually run, so no new dependency is needed. We only need to add flask or its extensions to the tox.ini testenv section and this allows us to actually run the whitelist file during testing.

I too think this is a good idea, and I think, it answers all the questions above.

@bje- can you add `pint` to the tox.ini testenv section? That should fix the CI and we should be good to go.

bje-:
Like so?

jendrikseipp (changes requested):
Looks good to me! Can you add a changelog entry, please?

jendrikseipp:
Thanks!

bje-:
And thank you!

PR #262, opened 2021-08-11
Add some pandas functions whitelist

mathbunnyru:
## Description
I added some pandas functions as a whitelist


jendrikseipp:
Thanks for the PR! There's one failing test for pypy3 on macos-10.15. I think the easiest solution is to skip testing the pandas whitelist for pypy3 on macos-10.15. Could you please look into how best to skip the test?

mathbunnyru:
Yes, I will do that couple of days later, I hope :)

Issue #47, opened 2017-06-05
unused functions' parameters and unused variables having same names are not reported.


RJ722:
For example. running vulture on the following file:
```python
first_name = "Rahul"
sur_name = "Jha"

def say_hello(first_name, sur_name):
	# first_name and sur_name are not reported
	pass

```
we would have following results:
```
dead.py:1: Unused variable 'first_name'
dead.py:2: Unused variable 'sur_name'
dead.py:4: Unused function 'say_hello'
```
It should also report that `first_name`, `sur_name` in `line 4` are not being used as in this case:
```python
first_name = "Rahul"
sur_name = "Jha"

def say_hello(a, b):
	# first_name and sur_name are not reported by vulture
	pass
```
which would produce:
```
dead2.py:1: Unused variable 'first_name'
dead2.py:2: Unused variable 'sur_name'
dead2.py:4: Unused function 'say_hello'
dead2.py:4: Unused variable 'a'
dead2.py:4: Unused variable 'b'
```

Is this a bug?

jendrikseipp:
This behaviour is to be expected since vulture doesn't distinguish between "local variables" and "parameter variables". We could actually distinguish between the "defined variable" and "defined parameter". Currently, vulture uses (in line 300)

        elif isinstance(node.ctx, (ast.Param, ast.Store)):
            self._define_variable(node.id, node.lineno)

We would need separate lists for "defined variables" and "defined parameters" and split the "elif" into two "elifs" for ast.Store and ast.Param. However, I don't see an easy way for distinguishing between "used variable" and "used parameter". Therefore, we can't detect unused parameters.

SmartManoj:
> dead2.py:4: Unused variable 'a'
> dead2.py:4: Unused variable 'b

Isn't a and b are unused params?

SmartManoj:
```
t.py:1: unused variable 'first_name' (60% confidence)
t.py:2: unused variable 'sur_name' (60% confidence)
t.py:5: unused function 'say_hello' (60% confidence)
t.py:5: unused parameter 'first_name' (100% confidence)
t.py:5: unused parameter 'sur_name' (100% confidence)
```

SmartManoj:
```
code = """
first_name = "Rahul"
sur_name = "Jha"


def say_hello(first_name, sur_name3):
    # first_name and sur_name are not reported
    print(first_name)
    pass
"""


import ast
import ast_scope
tree = ast.parse(code)
from pprint import *
[...]

PR #277, opened 2022-04-03
distinguishing var and param

SmartManoj:
## Related Issue
 #47 
## Checklist:

- [ ] I have run `tox -e fix-style` to format my code and checked the result with `tox -e style`.


SmartManoj:
vulture core.py  --min-confidence 100

core.py:81: unused parameter 'filename in _ignore_variable()' (100% confidence)

jendrikseipp:
Adding scopes to Vulture is a rather fundamental change and something I won't be able to look into in the near future.

Issue #278, opened 2022-04-21
Mark deleted variables as used


jendrikseipp:
Hello guys, for overridden functions and unused arguments, can the `vulture` follow the same rules as pylint and other tools.

It is described here: https://stackoverflow.com/a/14836005

And this is specified as the suggested solution by PyLint: http://pylint-messages.wikidot.com/messages:w0613

Currently I am trying to use `del` but `vulture` is still reporting the variable as `unused`

_Originally posted by @sshishov in https://github.com/jendrikseipp/vulture/issues/235#issuecomment-1105162434_

jendrikseipp:
Thanks for the report @sshishov! I agree that `del a` should mark the `a` variable as used.

Here is a little test case for this:

```
def foo(a):
    del a
    return 1

foo(2)
```

To implement this, one would need to count `Name(id='a', ctx=Del())` in the abstract syntax tree as a usage of variable `a`. Pull requests welcome :-)

PR #279, opened 2022-04-22
Add support for `del` keyword for overridden functions

sshishov:
Currently if you override the function of any framework or any library and you are not using all parameters, then the parameters which are not used are identified as `unused`. It is good practice while overriding function/method to keep the signature of it intact. Otherwise you can get potentially some issues with next releases etc.

## Description
Adding support for `del` operation inside the function. The keyword `del` will mark the variable as used now and we can safely implement the suggested workaround.

## Related Issue
This PR is addressing the issue #278 

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 * Print absolute filepaths as relative again (as with 2.1 and before) if they
   are below the current directory (The-Compiler, #246).
 * Run tests and add PyPI trove for Python 3.10 (chayim, #266).
-* Add support for `del` keyword to mark variable as used for function/methods overrides (sshishov, #279)
+* Allow using the `del` keyword to mark unused variables (sshishov, #279).
 
 # 2.3 (2021-01-16)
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 * Print absolute filepaths as relative again (as with 2.1 and before) if they
   are below the current directory (The-Compiler, #246).
 * Run tests and add PyPI trove for Python 3.10 (chayim, #266).
+* Add support for `del` keyword to mark variable as used for function/methods overrides
```
jendrikseipp, line 6: Do you want to add your name and the PR number here?

jendrikseipp (changes requested):
Thanks for the PR! Could you also describe this new recommendation for handling unused function arguments and its benefits in the README, please?

CHANGELOG.md
sshishov: Added

sshishov:
@jendrikseipp I have updated the README file. Could you please check? If you have better wording, please provide here, I will kindly make the changes.

jendrikseipp:
Thanks for the revisions!

PR #281, opened 2022-05-27
Pass file names by default

### LINKED ISSUE #280: passing filenames to pre-commit hook
I tried to use `vulture` as part of our `.pre-commmit-config.yaml` file. I added the following snippet directly as documented in [README](https://github.com/jendrikseipp/vulture/blob/master/README.md#version-control-integration):

```yaml
repos:
  - repo: https://github.com/jendrikseipp/vulture
    rev: 'v2.3'  # or any later Vulture version
    hooks:
      - id: vulture
```

It failed with the following error:

> Please pass at least one file or directory

While I fixed it by passing `.` as an argument in `args` section or by using `pass_filenames: true`, I feel this is a general enough to be included as part of the hook definition. Can you please change this?

https://github.com/jendrikseipp/vulture/blob/master/.pre-commit-hooks.yaml#L7

Or, if there are any issues with that, is it possible to at least update the documentation with a working snippet? I think that will be very helpful for other users.

yarnabrina:
## Description

Current hook definition stops passing file names as detected by `pre-commit`, and completely relies on the specifications in `pyproject.toml`.

Even though it's a good practice and quite common in python packages, I'd argue that end users (python developers using these packages) seldom configure with `pyproject.toml`. Their objective is just to run a tool to clean up the codebase as quickly as possible.

The changes in fork tries to resolve this by passing files from `pre-commit` by default (which is also the default for other standard linters - `pylint`, `flake8`, etc.).

If someone does configure vulture using `pyproject.toml`, they can still choose to use the current settings by defining paths in `pyproject.toml` and setting `pass_filenames` as `false`.

## Related Issue

resolves #280 

## Checklist:

- [x] I have updated the documentation in the README.md file.
- [ ] I have run `tox -e fix-style` to format my code and checked the result with `tox -e style`.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,6 +4,4 @@
   entry: vulture
   description: Find unused Python code.
   types: [python]
-  pass_filenames: false
   require_serial: true
-  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# unreleased
+
+* pass file names to `pre-commit` hook by default
+
 # 2.4 (2022-05-19)
 
 * Print absolute filepaths as relative again (as in version 2.1 and before)
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ following to the `.pre-commit-config.yaml` file in your repository:
 ```yaml
 repos:
   - repo: https://github.com/jendrikseipp/vulture
-    rev: 'v2.3'  # or any later Vulture version
+    rev: 'v2.4'  # or any later Vulture version
     hooks:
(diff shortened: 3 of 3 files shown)


jendrikseipp:
I read up on this issue again and found the reason why things are configured the way they are: with `pass_filenames: true` pre-commit will only check modified files and therefore produce many false positives (see https://github.com/jendrikseipp/vulture/pull/244). While researching this, I found `pre-commit run --all-files` but I think the current solution of requiring the user to specify the files leads to better results.

I don't see a better solution than the status quo, so I'm closing this PR. But we can definitely reopen it when we find a better solution.

Issue #172, opened 2019-09-14
Mark imports in `__all__` as used


Zac-HD:
I've been experimenting with Vulture on [Hypothesis](https://github.com/HypothesisWorks/hypothesis/tree/master/hypothesis-python/src/hypothesis) - which has found an unused argument I'd like to fix, so thanks! - and noticed that *some* of the imports to [this module](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/strategies.py) are flagged as unused with 90% confidence.

Of course, the whole point of that module is just to make things publicly available, and we verify that at runtime at the bottom of the module.  Only four names make it into the Vulture output, so I assume that `__all__` is supposed to be taken into account and it's just failing here for some reason?

jendrikseipp:
So far, we don't take `__all__` into account, but I think this feature should be added. I'll leave this issue open to address that. Pull requests welcome :-)

In your case, it seems some of your imports are not mentioned in other parts of the analyzed code.

jendrikseipp:
@kreathon fixed this in #282.

PR #282, opened 2022-06-30
Mark imports in __all__ as used

kreathon:
At the moment, symbols that are referenced in string form in the `__all__`  variable assignment are not marked as used.

#### Example

```
# Currently, 'Bar' would be marked as unused.
from foo import Bar

__all__ = ["Bar"]

```

## Implementation

Add `visit_Assign` to cover assignments of the following form:

```
# Default case
__all__ = ["A"]

# Tuples are also tolerated
__all__ = ("B", )
```

#### Limitations

The implementation cannot cover dynamic definitions e.g.:

```
a = ["A"]
__all__ = a
```

## Related Issue
This PR is addressing the issue https://github.com/jendrikseipp/vulture/issues/172.

## Checklist:

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -630,10 +630,8 @@ class Vulture(ast.NodeVisitor):
         if _assigns_special_variable__all__(node):
             assert isinstance(node.value, (ast.List, ast.Tuple))
             for elt in node.value.elts:
-                if isinstance(elt, ast.Constant) and isinstance(
-                    elt.value, str
-                ):
-                    self.used_names.add(elt.value)
+                if isinstance(elt, ast.Str):
+                    self.used_names.add(elt.s)
 
     def visit_While(self, node):
         self._handle_conditional_node(node, "while")
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -53,6 +53,16 @@ def _is_test_file(filename):
     )
 
 
+def _assigns_special_variable__all__(node):
+    assert isinstance(node, ast.Assign)
+    name_targets = {
+        target.id for target in node.targets if isinstance(target, ast.Name)
+    }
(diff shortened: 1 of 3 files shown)


vulture/core.py
```diff
@@ -53,6 +53,16 @@ def _is_test_file(filename):
...
 
 
+def _assigns_special_variable__all__(node):
+    assert isinstance(node, ast.Assign)
+    name_targets = {
+        target.id for target in node.targets if isinstance(target, ast.Name)
+    }
+    return "__all__" in name_targets and isinstance(
+        node.value, (ast.List, ast.Tuple)
+    )
```
jendrikseipp, line 63: I can imagine that the following versions runs slightly faster, since it avoids a temporary set and makes the cheap if-test first:

```suggestion
def _assigns_special_variable__all__(node):
    assert isinstance(node, ast.Assign)
    return (
        isinstance(node.value, (ast.List, ast.Tuple)) and
        any(target.id == "__all__" for target in node.targets if isinstance(target, ast.Name))
    )
```

jendrikseipp (changes requested):
Code looks great, thanks! I only have one suggestion, see below. And I was wondering if we should add more tests to cover edge cases such as ignoring code like the following `import my_module; my_module.__all__ = ["Foo"]`. What do you think?

vulture/core.py
kreathon: I agree that it runs faster, but in my opinion it is slightly less readable.

But of course I am fine with it ;)

kreathon:
Thank you for your feedback. 

I am  going to add a few more test cases (also to increase the branch coverage).

jendrikseipp:
I agree about the readability, but let's go with that version anyway. Thanks for your great work!

PR #284, opened 2022-07-10
Decorator name regression

jendrikseipp:


vulture/noqa.py
```diff
@@ -3,7 +3,7 @@
 
 NOQA_REGEXP = re.compile(
     # Use the same regex as flake8 does.
-    # https://gitlab.com/pycqa/flake8/-/tree/master/src/flake8/defaults.py
+    # https://gitlab.com/pycqa/flake8/-/tree/main/src/flake8/defaults.py
```
RJ722, line 6: `flake8` has moved to GitHub in the meantime:

```suggestion
    # https://github.com/PyCQA/flake8/blob/main/src/flake8/defaults.py
```

RJ722: Oops, looks like these changes are only appearing here because GitHub wasn't able to rebase this branch (probably because of the `master` -> `main` rename). Changes already went in fb126c7c24e54e737bcadce9cc16add9413780c6. I'll send a PR separately.

jendrikseipp:
This PR is probably redundant because #285 tackles the same problem.

Issue #283, opened 2022-07-05
AttributeError when parsing a decorator


Llandy3d:
Hello,

I just wanted to raise awareness about an issue that I mainly encountered when working with the python prometheus client.

The `get_decorator_name` function seems to assume that after cycling over the `ast.Attribute` you can get the id `decorator.id`.
When using a decorator from the `prometheus_client` for example:
```python
from prometheus_client import Histogram
hist = Histogram('name', 'description', labelnames=["label1"])

@hist.labels('place1').time()
def myfunc():
    ...
```
after the Attribute you will be finding an `ast.Call` object that won't have the id attribute and raise the `AttributeError`. From my understanding this is due to some extra decorating work that the library is doing, but possibly the `vulture` tool shouldn't fail with an error.

For now this has been solved by ignoring the file with the `--exclude` flag

The traceback:
```python
Traceback (most recent call last):
  File "/usr/local/bin/vulture", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 689, in main
    vulture.scavenge(config["paths"], exclude=config["exclude"])
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 264, in scavenge
    self.scan(module_string, filename=module)
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 231, in scan
    self.visit(node)
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 645, in visit
    return self.generic_visit(node)
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 677, in generic_visit
    self.visit(item)
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 630, in visit
    visitor(node)
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 564, in visit_FunctionDef
    decorator_names = [
  File "/usr/local/lib/python3.10/site-packages/vulture/core.py", line 565, in <listcomp>
    utils.get_decorator_name(decorator)
  File "/usr/local/lib/python3.10/site-packages/vulture/utils.py", line 69, in get_decorator_name
    return "@" + ".".join(reversed(parts))
AttributeError: 'Call' object has no attribute 'id'
```

jendrikseipp:
[...]

PR #285, opened 2022-07-10
Fix get_decorator_name with callable in between

Llandy3d:
## Description

`get_decorator_name` would fail with an `AttributeError` on decorators on the style of the `prometheus_client` where you actually have a call in between instead of just at the end of it. Ex. `@hist.labels('label').time()`

This pr changes the method to loop so that we will catch all the `ast.Call` and can build the full name.

## Related Issue

https://github.com/jendrikseipp/vulture/issues/283

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.6 (Unreleased)
+
+* Fix `get_decorator_name` regression (Llandy3d, #285).
+
 # 2.5 (2022-07-03)
 
 * Mark imports in `__all__` as used (kreathon, #172, #282).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 2.6 (Unreleased)
 
-* Fix `get_decorator_name` regression (Llandy3d, #285).
+* Fix `get_decorator_name` with callable in between (Llandy3d, #285).
 
 # 2.5 (2022-07-03)
 
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,6 +1,7 @@
 import ast
 import os
 import pathlib
+import sys
 
 import pytest
 
@@ -121,6 +122,9 @@ class Foo:
     check_decorator_names(code, ["@foo", "@bar.yz"])
 
 
+@pytest.mark.skipif(
+    sys.version_info < (3, 9), reason="requires python3.9 or higher"
+)
 def test_get_decorator_name_regression():
     code = """\
 from prometheus_client import Histogram
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -119,3 +119,14 @@ class Foo:
     pass
 """
     check_decorator_names(code, ["@foo", "@bar.yz"])
+
+
+def test_get_decorator_name_regression():
+    code = """\
+from prometheus_client import Histogram
+hist = Histogram("name", "description", labelsname=["label1"])
+@hist.labels("labelvalue").time()
+def myfunc():
+    pass
+"""
+    check_decorator_names(code, ["@hist.labels.time"])
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -125,12 +125,37 @@ class Foo:
 @pytest.mark.skipif(
     sys.version_info < (3, 9), reason="requires python3.9 or higher"
 )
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 2.6 (Unreleased)
+
+* Fix `get_decorator_name` regression (Llandy3d, #285).
```
Llandy3d, line 3: I saw a mix of ways to add to the changelog, I hope this one is ok!

tests/test_utils.py
```diff
@@ -119,3 +120,17 @@ class Foo:
     pass
 """
     check_decorator_names(code, ["@foo", "@bar.yz"])
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 9), reason="requires python3.9 or higher"
+)
```
Llandy3d, line 127: the only thing problematic for older versions is the test, so we ignore this specific test

vulture/utils.py
```diff
@@ -54,12 +54,15 @@ def format_path(path):
...
-        parts.append(decorator.attr)
-        decorator = decorator.value
+    while True:
+        if isinstance(decorator, ast.Call):
+            decorator = decorator.func
+        while isinstance(decorator, ast.Attribute):
+            parts.append(decorator.attr)
+            decorator = decorator.value
+        if not isinstance(decorator, ast.Call):
+            break
```
Llandy3d, line 65: in the regression test this will return something in the lines of `@hist.labels.time` I'm not completely sure if this is what is wanted, but atleast now in that specific case it's not failing 🤔

RJ722: The other option would be to return calls with `args`/`kwargs` preserved (like `@hist.labels("labelvalue").time`).

For context, the reason why `get_decorator_method` exists is for us to be able to support the `--ignore-decorators` flag. Preserving `args`/`kwargs` here would make it possible for users to ignore for particular arguments/kwargs only (e.g. `--ignore-decorators "@hist.labels("labelA").time"`), but I'm not sure it's possible to do that statically. Besides, even if it is, it would require significant changes to the code and I don't think it worth the maintenance effort. If a user wants to ignore for one value of `labelValue`, they probably want to do so for other values too.

RJ722 (commented):
Thanks for the PR, @Llandy3d!

Other than the few minor comments, I think we should mention this behavior (how to specify a decorator with a callable in between to `--ignore-decorators`) in the docs as well. Perhaps, add an example in the README?

jendrikseipp:
Thanks for making these additional changes @Llandy3d ! I'll have a look at the PR when @RJ722 is happy :-)

Llandy3d:
@jendrikseipp just pinging as I'm not sure if github gives notifications for emoji responses, but seems like a positive one? 😁

RJ722:
Hi @Llandy3d! I appreciate the reminder -- It comes at just the right time, when I have found some time to work on open source. I'll give another round of review in a few hours. ^>^. Thanks for the patience!

Llandy3d:
Hi @RJ722 , I probably misunderstood the hooray icon response as a positive review 😅
No rush at all, and thank you!

jendrikseipp:
First off, thanks for looking into this, @Llandy3d and @RJ722 ! The code fixes the issue you were seeing, but I think we need a more general solution to cover the flexible decorators that Python 3.9 allows: a decorator can be any expression now (https://peps.python.org/pep-0614/). For example `@foo[2]` is valid, but not accepted by the code in this PR.

Since we cannot hope to cover all possible decorator names, I suggest we use the following basic solution: 
* `utils.get_decorator_name()` preserves its old behaviour but returns the empty string for decorator names that it cannot parse (caught in an `except AttributeError` block) 
* we update the readme file to briefly reflect that we only support decorators of the form @foo.bar(x, y, z)

@Llandy3d do you want to incorporate these changes into this PR?

jendrikseipp:
@Llandy3d If you don't have the time to continue the PR that's completely fine of course. If so, maybe @RJ722 or I can pick up where you left off.

Llandy3d:
sorry I missed the first message. TIL about that on decorators :)
it makes sense, I don't know if/when I would be able to pick this back up so please feel free to work on it! 🙇‍♂️

jendrikseipp:
I'm making the fix discussed above in #284.

Issue #276, opened 2022-03-15
vulture does not detect code usage by `case` clauses in pattern matching


exoriente:
Hi,

I'm running into a issue with pattern matching and vulture. In my case the property of a class is only used during pattern matching. Vulture doesn't pick up on it and considers it dead code. I've whitelisted the properties for now, but it would be great if vulture could be improved to cover this new scenario.

Pseudo code example:

```python
@dataclass
class MyClass
    my_field: MyType
    
    @property
    my_test(self) -> bool:
        return my_field == my_constant

def do_something(x: Object) -> None:
    match x:
        case MyClass(my_test=True):
            do_stuff(x)
        case MyClass(my_test=False):
            do_other_stuff(x)
```

In this example, vulture seems to flag `my_test` as dead code.

jendrikseipp:
Thanks for the report! We haven't added explicit support for the `match` statement, yet, so there could be room for improvement.

But in your code, I think `my_test` is really unused. Shouldn't the `my_test={True,False}` kwarg be called `my_field`?

PR #291, opened 2022-09-16
Basic match statement support

kreathon:
#### Example

```
@dataclass
class X:
    a: int

x = input()

match x:
    case X(a=0):
        ...
    case _:
        ...
```

At the moment, `a` would be seen as unused. However, as nicely described by @exoriente, the code means something like

```
    if is_instance(x, X) and x.a == 0:
```

So it is not a normal / dead "assignment" and `a` should not be seen as unused.

## Implementation
The actual code change is really trivial, since most of the desired behavior (for the match statement) is already handled by `visit_Name`:

```
    def visit_MatchClass(self, node):
        for kwd_attr in node.kwd_attrs:
            self.used_names.add(kwd_attr)
```

#### Limitations
- cannot handle positional arguments (can be tricky since overwrites of `__match_args__` are possible)
- cannot handle variable definitions in `case`: e.g. "case [1, a, *rest]" (`a` and `rest` is not handled i.e. checked if it is used in the case body). Currently, I do not see any problem implementing it in the future.

#### Python Compatibility

For Python < 3.10 the `visit_MatchClass` should never be executed. Tests are guarded with `@pytest.mark.skipif(
    sys.version_info < (3, 10), ...)`

## Related Issue
This PR is addressing the issue https://github.com/jendrikseipp/vulture/issues/276.

## Checklist:

### DIFF
--- a/tests/test_scavenging.py
+++ b/tests/test_scavenging.py
@@ -1,5 +1,7 @@
 import sys
 
+import pytest
+
 from . import check, v
 
 assert v  # Silence pyflakes.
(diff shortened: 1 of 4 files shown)


tests/test_scavenging.py
```diff
@@ -759,3 +761,99 @@ def foo(a, b, c, d=3):
...
+        print("Real danger!")
+    case Color.YELLOW | Color.GREEN:
+        print("No danger!")
+"""
+    )
+    check(v.defined_classes, ["Color"])
+    check(v.defined_vars, ["RED", "YELLOW", "GREEN", "color"])
+
+    check(v.unused_classes, [])
+    check(v.unused_vars, [])
```
jendrikseipp, line 859: Nice tests! Can you add one unused variable in each case, please? That is, a dataclass member that's not used in the match statement.

jendrikseipp (changes requested):
Thanks for this very polished PR! I only have one small comment, see below.

tests/test_scavenging.py
kreathon: Thank you, I updated the test cases.

jendrikseipp:
Great! Thanks @kreathon !

exoriente:
:star_struck: Awesome! Nice to see this addressed! Thanks, for taking up this issue @jendrikseipp & @kreathon !

PR #293, opened 2022-10-12
Whitelist certain socketserver.TCPServer attributes

bje-:
Seems like a reasonable change given that `socketserver` is in the Python standard library.


jendrikseipp:
Thanks!

PR #294, opened 2022-11-10
Adding a new option --exit-zero

liyao001:
Related tools, like pylint and flake8, have similar options to make the program always return zero so that the tool can be easily integrated into continuous integration pipelines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,9 @@ codes.
 |     1     | Invalid input (file missing, syntax error, wrong encoding)    |
 |     2     | Invalid command line arguments                                |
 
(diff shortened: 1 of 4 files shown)


RJ722:
Hello @liyao001, one way to introduce Vulture into your CI/CD pipelines without relying on the `--exit-zero` flag is to call it with `|| true` at the end:

```bash
$ vulture . || true
```

I'm a bit opposed to adding functionality (which we'll need to test and maintain) if the same can be achieved easily by other means.

jendrikseipp:
I agree with @RJ722.

Issue #299, opened 2023-01-16
UnicodeEncodeError in core.py


milanbalazs:
I got `UnicodeEncodeError` exception during the `Vulture` analysis.

**Traceback:**

```
Traceback (most recent call last):
    File "/ws/venv3_x/bin/vulture", line 8, in <module>
      sys.exit(main())
    File "/ws/venv3_x/lib/python3.6/site-packages/vulture/core.py", line 710, in main
      make_whitelist=config["make_whitelist"],
    File "/ws/venv3_x/lib/python3.6/site-packages/vulture/core.py", line 334, in report
      else item.get_report(add_size=sort_by_size)
  UnicodeEncodeError: 'latin-1' codec can't encode characters in position 84-90: ordinal not in range(256)
```

**My command:**

```
vulture "--exclude" "venv2/,venv3_*,*update_repo.py" "."
```

**Used `vulture` version in `requirements.txt file`:**

 - `vulture==2.5`

**Used `Python` version:**
 - `3.6.6`

It's possible that my codebase contains special characters but it's hard to debug it because I analyze ~1000 Python files in the same time (And the provided Exception doesn't contain any reference for the problematic file).

I think it would be nice to improve the Encoding in the `core.py` or at least handle the exception and throw a more talkative exception (perhaps with the name/path of the problematic file)

Thank you in advance!

milanbalazs:
I have changed the `venv3_*` pattern to `*venv3_*` in the `--exclude` parameter and now the analysis works as expected. I have removed all `non-ascii` characters from my code-base so in my opinion the faulty file is in the Virtual Environment.

On the other hand, I still think the Exception should me more specific/talkative and the error should be handled.

jendrikseipp:
Thanks for the report! I agree that the error message should include the file name. Would you care to prepare a PR for this?

milanbalazs:
Sure, I will create a PR for it!

PR #300, opened 2023-01-18
Add `UnicodeEncodeError` exception handling to `core.py`

milanbalazs:
## Description

If the code contains special characters which are not part of the `latin-1` codec, the `print` function will be failed with an `UnicodeEncodeError` exception. Like below: 

```
UnicodeEncodeError: 'latin-1' codec can't encode characters in position 84-90: ordinal not in range(256)
```

If the string is encoded (default is `utf-8`) then the print function is able to print the related string. The users can identify the related file and line.

**An example output with my change:**

```
b"test.py:6: unused variable '\xd0\xb2\xd0\x82\xd0\x87' (60% confidence)"
```

**Note:**

 - My change doesn't change the original behavior and it's totally backward compatible. 

## Related Issue

- https://github.com/jendrikseipp/vulture/issues/299

## Checklist:

- [✅] I have updated the documentation in the README.md file or my changes don't require an update.
- [✅] I have added an entry in CHANGELOG.md.
- [✅] I have added or adapted tests to cover my changes.
- [✅] I have run `tox -e fix-style` to format my code and checked the result with `tox -e style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Add `UnicodeEncodeError` exception handling to `core.py` (milanbalazs, #299).
+
 # 2.7 (2023-01-08)
 
 * Ignore `setup_module()`, `teardown_module()`, etc. in pytest `test_*.py` files (Jendrik Seipp).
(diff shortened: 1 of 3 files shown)


jendrikseipp:
Elegant solution! Thanks!

Issue #270, opened 2022-01-03
Should report on unreachable code after try-return-except-raise


kretes:
For a following file:
```
def test() -> int:
    try:
        return 2
    except Exception as e:
        raise e
    return 1
test()
```

`vulture dead_return.py`
doesn't report anything, while it should report code on line 6, as this one is unreachable.

vulture 2.3, python 3.9

jendrikseipp:
Thanks for the report! Put a little more generally, Vulture should detect that code is reachable if it occurs after a if-else block or a try-except block where all cases exit the function (with `return` or `raise`).

Here is another test case that Vulture currently misses:

```
def test2(a) -> int:
    if a:
        return 1
    else:
        return 2
    return 3
test2(5)
```

I'm happy to review a PRs for this :-)

adamchainz:
Mypy has good unreachable code detection, and it can work for such cases even if you don’t use type hints. https://adamj.eu/tech/2021/05/19/python-type-hints-mypy-unreachable-code-detection/

I am not sure it’s worth duplicating this kind of functionality in vulture.

kykyi:
I disagree, mostly because this is dead-code and users of vulture are expecting it to find dead-code 🤷

PR #302, opened 2023-02-17
Improve unreachable code analysis

kreathon:
#### Motivation

```
def f():
    try:
        return
    except:
        raise Exception()
    print("Unreachable")
```

At the moment, the `print` statement is not identified as unreachable.

## Implementation
I added a data structure (`no_fall_through_nodes`) that stores ast nodes that do not allow a fall through. During the ast traversal `continue`, `break`, `raise` and `return` are added into the this data structure. For every control flow statement / object (`for`, `try`, `while`, `with`, but also `module` and `functiondef`), we check if any of the statement in the body is in `no_fall_through_nodes`. If that is the case, we report the next statement (if there is any) and add the current node also into `no_fall_through_nodes`. To make this work `generic_visit` (means visiting children) is now executed before visiting the current node.

The algorithm was added into the existing `Vulture` object with the `generic_visit` (which handles recursion). I think a cleaner implementation could implement a separate node visitor (that handles its own recursion), but I was not sure if the project is open for that.

The old error reporting message is reused (e.g. `unreachable code after 'try'`). I am not sure if this is the best message or if all unreachable messages should be simplified to something like `unreachable code`.

## Limitations
- does not support `match` statements
- does not support `with` statements  (context manager can suppress exceptions)

## Related Issue
This PR is addressing the issue https://github.com/jendrikseipp/vulture/issues/270.

## Checklist:


jendrikseipp (changes requested):
Thanks for raising this PR! I like the algorithm and that it nicely generalizes the earlier solution. However, I'd like to propose a different code organization, so that the core module doesn't grow too large:

* Add a `reachability.py` module with a `Reachability` class.
* This class stores the no-fall-through-nodes, allows to reset them, has a `visit(node)` method that checks the node's type and depending on the type marks nodes as no-fall-through or checks whether their body is fully executable.
* The Vulture class has a Reachability member and calls its `visit(node)` function, and retrieves the unreachable code from Reachability before switching to the next module.

This goes in the direction of adding a second NodeVisitor class, but avoids traversing the AST twice. Or do you think a second NodeVisitor class (with all of these visitor_* functions) would be a better alternative?

kreathon:
> This goes in the direction of adding a second NodeVisitor class, but avoids traversing the AST twice. Or do you think a second NodeVisitor class (with all of these visitor_* functions) would be a better alternative?

I do not see a clear advantage of one of the solutions and it probably does not matter too much right now (if the reachability is in a second module it should be really easy to switch between them anyway).

I will update the code according to your proposed code organization.

kreathon:
I am not sure if passing `_define` as `report` into `Reachaiblity` is the cleanest solution, but I did not want to create a new datastructure that passes the information back to the `Vulture` object. What do you think?

kreathon:
I had a look at it again, and I think it is cleaner to also move the `_handle_conditional_node` into the `reachablitly.py` (such that the module handles the entire `self.unreachable_code` `LoggingList` data structure).

I will propose the change as soon as I find time for it.

kreathon:
I updated

-  `check_unreachable` for checking multiple unreachable code segments (alternative would be to add another method for that or assert that there is only a single unreachable code segment separately)
- move some tests from `test_conditions.py` into `test_reachability.py`
- merge `handle_condition_node()` into `Reachability` to get more powerful analysis results. There are still cases that are not handled (we need to distinguish between raise/return and break/continue for that).

For example.
```
while True:
    raise Exception()
print(b)
```

I could handle this in another PR.

jendrikseipp (approved):
This flew under my radar, sorry. I had three minor comments, but took care of them myself now. Thanks for your work on this, it not only generalizes the code, but also makes it nicer!

jendrikseipp:
>  There are still cases that are not handled (we need to distinguish between raise/return and break/continue for that).
> 
> 
> For example.
> 
> ```
> while True:
>     raise Exception()
> print(b)
> ```
> 
> I could handle this in another PR.

Feel free to do so, if you're still interested :)

kreathon:
> This flew under my radar, sorry. I had three minor comments, but took care of them myself now. Thanks for your work on this, it not only generalizes the code, but also makes it nicer!

I actually thought recently about this PR again and if I should rebase and give it another go 😅

PR #306, opened 2023-02-25
Add special enum attributes to whitelist.

### LINKED ISSUE #305: False positives on enum `_name_` and `_value_` attributes
```python
from enum import Enum

class E(Enum):
    A = ('a', 1)
    B = ('b', 2)

    def __init__(self, name, value):
        self._name_ = name
        self._value_ = value

E.A, E.B
```
Here vulture says that attributes `_name_` and `_value_` are unused:
```
$ python -m vulture enum_sunder.py
enum_sunder.py:8: unused attribute '_name_' (60% confidence)
enum_sunder.py:9: unused attribute '_value_' (60% confidence)
```
These are special attributes used to set enum members' name and value respectively. They should be whitelisted.

eltoder:
Closes #305

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 * Add `UnicodeEncodeError` exception handling to `core.py` (milanbalazs, #299).
+* Add whitelist for `Enum` attributes `_name_` and `_value_` (Eugene Toder, #305).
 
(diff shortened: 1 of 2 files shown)


jendrikseipp:
Great, thanks!

PR #310, opened 2023-04-04
Add the ability to use the full path for exclude folder/file

mirecl:
## Description
**AS-IS:**
```bash
vulture . --exclude=scripts,app.py  # run command
```
*Vulture* exclude all folders/files with match `*scripts*` or `*app.py*` and no way to specify the exact location of the object.

**TO-BE:**
```bash
vulture . --exclude=./scripts,./app.py  # run command
```
Now *vulture* adding `./ ` to exclude will determine the full path to the object:
```python
['/Users/<name>/Desktop/folder/scripts/*', '/Users/<name>/Desktop/folder/app.py'
```

## Checklist:

### DIFF
--- a/vulture/core.py
+++ b/vulture/core.py
@@ -2,6 +2,7 @@ import ast
 from fnmatch import fnmatch, fnmatchcase
 from pathlib import Path
 import pkgutil
+import os
 import re
 import string
 import sys
@@ -261,6 +262,11 @@ class Vulture(ast.NodeVisitor):
 
     def scavenge(self, paths, exclude=None):
         def prepare_pattern(pattern):
+            if pattern.startswith('./') is True:
+                pattern = os.path.abspath(pattern)
+                if os.path.isfile(pattern) is False:
+                    pattern = f"{pattern}/*"
+                return pattern
             if not any(char in pattern for char in "*?["):
                 pattern = f"*{pattern}*"
             return pattern


mirecl:
@jendrikseipp, what do you think? will it be useful for the project?

jendrikseipp:
Thanks for the PR! It'll take me a while until I can review it properly. Two quick notes:

* When working on the ignore mechanism, we don't want to break backwards compatibility.
* Ideally, the ignore mechanism would be similar to other popular tools such as black and flake8.

If you can rework the PR in these directions, that'll help getting it merged. If you already check both of these, all the better :-)

mirecl:
@jendrikseipp, since this is a new feature - backwards compatibility should remain in full for option exclude files.

Option exclude files with ``./ `` must not conflict with **flake8** and **black**. But the current option exclude files approach definitely contradict **flake8** and **black**:
https://github.com/jendrikseipp/vulture/blob/e2e84d0a63201742ccee4bfcbe7ab065ed100f2c/vulture/core.py#L263-L266

mirecl:
@jendrikseipp, pls tell us how the progress on this PR is going:)

mirecl:
@jendrikseipp, what is the status of PR?

jendrikseipp (changes requested):
Thanks for your work on this! Please also add a short changelog entry.

jendrikseipp:
@mirecl Are you still planning to work on this?

PR #313, opened 2023-05-07
Fix false positives in typing.Protocol

### LINKED ISSUE #309: Using with typing.Protocol
Suppose I have protocol class:
```python
from typing import Protocol


class LoggerProtocol(Protocol):
    def log(
        self, level: int, msg: str, *args, extra: Mapping[str, object] | None = None
    ) -> None:
        ...
```
After execution of vulture I get the following output:
```
logging.py:24: unused variable 'args' (100% confidence)
logging.py:24: unused variable 'extra' (100% confidence)
logging.py:24: unused variable 'msg' (100% confidence)
```

In my opinion, vulture should ignore this type of "confidence", due to the nature of typing.Protocol

pm3512:
## Description

Adds the `ast_scope` package and uses scoping information to find functions in subclasses of `typing.Protocol` and mark them and their arguments as used.
## Related Issue

Resolves #309

## Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install --upgrade build coveralls setuptools tox==4.0.2 wheel ast-scope=0.3.1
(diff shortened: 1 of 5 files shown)


btschroer:
Nice 👍

Issue #327, opened 2023-09-25
Make whitelist configurable


mssalvatore:
I'd like to be able to configure the path to my vulture whitelist in pyproject.toml. At the present time, it appears that vulture must be invoked like `vulture TARGET WHITELIST_FILE` if you'd like items in the whitelist to be ignored. I'd like to be able to configure `path` and `whitelist` in my pyproject.toml so that I can simply invoke `vulture`.

Thanks!

jendrikseipp:
You can set the `paths` variable in `pyproject.toml` to achieve this: `paths = [TARGET, WHITELIST_FILE]`

mssalvatore:
Thanks!

PR #328, opened 2023-09-26
Whitelist path documentation

mssalvatore:
## Description
Adds explicit documentation regarding how the whitelist file can be configured in the `pyproject.toml`.

## Related Issue
Issue #327

## Checklist:

- [ ] ~I have added an entry in CHANGELOG.md~. N/A
- [ ] ~I have added or adapted tests to cover my changes~. N/A


jendrikseipp:
Thanks for the PR! I think having the extended example is enough, so I removed the extra text.

PR #340, opened 2023-11-23
Swap in tomllib/tomli for toml.

### LINKED ISSUE #339: Switch to tomli/tomllib for toml parsing
Due to https://github.com/uiri/toml/issues/270, vulture won't run for projects with a `pyproject.toml` that contain an array with mixed types. According to https://github.com/python-poetry/poetry/issues/7094#issuecomment-1328127456, this is not an issue in the `tomllib` in Python 3.11+ or its backport (https://github.com/hukkin/tomli).

It looks like this should be a quick swap. I'll send a PR along when I have a moment.

sacsar:
## Description

Fixes #339 -- the toml package does not support heterogenous arrays, which are now allowed in the toml spec. This commit swaps in tomllib from the standard library for Python >= 3.11 and its backport tomli for older versions.

## Checklist:
- [X] I have updated the documentation in the README.md file or my changes don't require an update.
- [X] I have added or adapted tests to cover my changes.
- [X] I have run `tox -e fix-style` to format my code and checked the result with `tox -e style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-# 2.11 (2023-11-13)
-* Switch to tomllib/tomli to support heterogenous arrays. (Sebastian Csar, #340)
+# next (unreleased)
+* Switch to tomllib/tomli to support heterogeneous arrays (Sebastian Csar, #340).
 
 # 2.10 (2023-10-06)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.11 (2023-11-13)
+* Switch to tomllib/tomli to support heterogenous arrays. (Sebastian Csar, #340)
+
 # 2.10 (2023-10-06)
 
 * Drop support for Python 3.7 (Jendrik Seipp, #323).
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -17,6 +17,13 @@ from vulture.config import (
 )
 
 
+def get_toml_bytes(toml_str: str) -> BytesIO:
+    """
+    Wrap a string in BytesIO to play the role of the incoming config stream.
+    """
+    return BytesIO(bytes(toml_str, "utf-8"))
+
+
 def test_cli_args():
     """
     Ensure that CLI arguments are converted to a config object.
@@ -62,10 +69,9 @@ def test_toml_config():
         sort_by_size=True,
         verbose=True,
     )
-    data = BytesIO(
-        bytes(
-            dedent(
-                """\
+    data = get_toml_bytes(
+        dedent(
+            """\
         [tool.vulture]
         exclude = ["file*.py", "dir/"]
         ignore_decorators = ["deco1", "deco2"]
@@ -76,8 +82,6 @@ def test_toml_config():
         verbose = true
         paths = ["path1", "path2"]
         """
-            ),
-            "utf-8",
         )
     )
     result = _parse_toml(data)
@@ -100,10 +104,9 @@ def test_toml_config_with_heterogenous_array():
         sort_by_size=True,
         verbose=True,
     )
-    data = BytesIO(
-        bytes(
-            dedent(
-                """\
+    data = get_toml_bytes(
+        dedent(
+            """\
         [tool.foo]
         # comment for good measure
         problem_array = [{a = 1}, [2,3,4], "foo"]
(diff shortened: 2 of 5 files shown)


sacsar:
Note: While the `tox` style check passes, `flake8` finds errors when run via `pre-commit` (including in `vulture/core.py`, which I didn't touch). I believe this is `flake8` and `black` fighting, but I didn't go too far down the rabbit hole and ended up using `--no-verify`.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.11 (2023-11-13)
+* Switch to tomllib/tomli to support heterogenous arrays. (Sebastian Csar, #340)
```
jendrikseipp, line 2:
```suggestion
* Switch to tomllib/tomli to support heterogeneous arrays (Sebastian Csar, #340).
```

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.11 (2023-11-13)
```
jendrikseipp, line 1:
```suggestion
# next (unreleased)
```

tests/test_config.py
```diff
@@ -75,6 +76,50 @@ def test_toml_config():
...
+        sort_by_size=True,
+        verbose=True,
+    )
+    data = BytesIO(
+        bytes(
+            dedent(
+                """\
+        [tool.foo]
+        # comment for good measure
+        problem_array = [{ a = 1}, [2,3,4], "foo"]
```
jendrikseipp, line 109:
```suggestion
        problem_array = [{a = 1}, [2,3,4], "foo"]
```

tests/test_config.py
```diff
@@ -87,9 +132,10 @@ def test_config_merging():
     If we have both CLI args and a ``pyproject.toml`` file, the CLI args should
     have precedence.
     """
-    toml = StringIO(
-        dedent(
-            """\
+    toml = BytesIO(
```
jendrikseipp, line 135: Could you factor this into a function, please?

```
def get_toml_bytes(config_string):
    return BytesIO(bytes(dedent(config_string), "utf-8"))
```

jendrikseipp (changes requested):
Thanks for the PR! Code looks good, I only have a few minor nitpicks.

tests/test_config.py
sacsar: I'm thinking of going with
```py
def get_toml_bytes(toml_str: str) -> BytesIO:
    """
    Wrap a string in BytesIO to play the role of the incoming config stream.
    """
    return BytesIO(bytes(toml_str, "utf-8"))
```
i.e. keeping `dedent` in the tests themselves. My thinking is that it then becomes clear why one is writing `"""\` rather than `"""`.

sacsar:
> Note: While the `tox` style check passes, `flake8` finds errors when run via `pre-commit` (including in `vulture/core.py`, which I didn't touch). I believe this is `flake8` and `black` fighting, but I didn't go too far down the rabbit hole and ended up using `--no-verify`.

It turns out this isn't a conflict with `black`, it's a change in `flake8` between Python versions. When I run `tox` from a 3.12 virtualenv, the style check fails with:
```
tests/test_utils.py:146:32: E272 multiple spaces before keyword
vulture/config.py:119:69: E231 missing whitespace after ','
vulture/config.py:120:25: E231 missing whitespace after ','
vulture/config.py:120:30: E231 missing whitespace after ','
vulture/config.py:120:42: E231 missing whitespace after ','
vulture/config.py:130:59: E231 missing whitespace after ','
vulture/config.py:138:71: E231 missing whitespace after ','
vulture/core.py:158:41: E231 missing whitespace after ':'
vulture/core.py:172:50: E231 missing whitespace after ':'
```
When run from a 3.8 virtualenv, flake8 is happy. I was getting failures only from pre-commit because I had installed `pre-commit` from the 3.12 virtualenv and then switched virtualenvs.

It turns out that if you use `pyenv` and have pre-commit (or other utilities) installed with a Python it manages, it turns into a pain in the butt, so I ended up installing `pre-commit` via `pipx` with my system python (3.11.5). It turns out there is a [bug in a recent version of flake8](https://stackoverflow.com/questions/77401175/how-to-make-flake8-ignore-syntax-within-strings) that was fixed in 6.1.0.

tl;dr @jendrikseipp I think the fix is just to bump flake8 in `tox.ini`. Do you want a separate PR in the name of a clean changelog or should I just throw it in here?

jendrikseipp:
Thanks for the changes and for looking into the flake8 issue! I'll merge this as is to have a clean history. Would be great if you could send another PR for the flake8 issue.

tests/test_config.py
jendrikseipp: Good catch!

PR #342, opened 2023-12-06
Provide whitelist parity for `MagicMock` and `Mock`

maxrake:
## Description

There are whitelist entries for the `return_value` and `side_effect`
attributes of the `unittest.mock.Mock` class. This change seeks to
provide parity for the `unittest.mock.MagicMock` class, which is just a
subclass of `Mock` but with default implementations of most of the magic
methods.

Standard library reference:
https://docs.python.org/3/library/unittest.mock.html#magic-mock

## Checklist:

- [x] ~I have added or adapted tests to cover my changes.~
  - No whitelist tests were found to update

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # next (unreleased)
-* Bump flake8, flake8-comprehensions and flake8-bugbear. (Sebastian Csar,
-#341).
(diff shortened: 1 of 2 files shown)


jendrikseipp:
Thanks!

PR #347, opened 2024-01-14
Use ruff for linting

tqa236:
## Description

Hello, 

This MR configures [ruff](https://github.com/astral-sh/ruff), an extremely fast Python linter and code formatter to `vulture`. `ruff` can potentially replace all the current linting and formatting tools, namely `black`, `flake8` and `pyupgrade`.

The current config in `pyproject.toml` is the default one, I put it there for ease of your review and customization. 

Let me know if this change is welcome. I can make further adjustments to the config to suit `vulture` style if needed.

## Checklist:

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,14 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4.1.1
+    - uses: actions/setup-python@v5
+    - uses: pre-commit/action@v3.0.0
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,5 +10,4 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4.1.1
-    # - uses: actions/setup-python@v5
     - uses: pre-commit/action@v3.0.0
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -10,5 +10,5 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4.1.1
-    - uses: actions/setup-python@v5
+    # - uses: actions/setup-python@v5
     - uses: pre-commit/action@v3.0.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ exclude = [
 ]
 
 # Same as Black.
-line-length = 88
+line-length = 79
 indent-width = 4
 
 # Assume Python 3.8
@@ -60,7 +60,7 @@ target-version = "py38"
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
-select = ["E4", "E7", "E9", "F"]
+select = ["E4", "E7", "E9", "F", "B"]
 ignore = []
 
 # Allow fix for all enabled rules (when `--fix`) is provided.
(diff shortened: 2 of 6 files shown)


jendrikseipp:
Thanks for the PR! I like the proposed changes. In addition, can you make sure that we enable the same ruff tests as were enabled previously via the other tools? And then remove the other tools?

tqa236:
Hi @jendrikseipp, this PR is ready for review.

There are a few things to consider:
- `ruff` can replace all existing linting packages. I have configured it to the same rules and removed the redundant packages.
- `ruff`  can even replace `black` with as a main formatter (some big packages like `pandas` already migrated, [ref](https://github.com/pandas-dev/pandas/pull/56704)). I can do this in the next PR if you think `vulture` can adopt this change. [This](https://docs.astral.sh/ruff/formatter/black/) is the documentation about the known deviations between `ruff` and `black`.
- I keep the current `tox -e style` and `tox -e fix-style` as the main entry points for now, but they can also be replaced completely with `pre-commit`. I do that in the next PR if you're ok with the deprecations of these `tox` commands.

pyproject.toml
```diff
@@ -17,3 +17,84 @@ exclude = '''
...
+    ".vscode",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "site-packages",
+    "venv",
+]
```
jendrikseipp, line 50: Can you keep only the excludes that are relevant for a typical Vulture checkout, please?

pyproject.toml
```diff
@@ -17,3 +17,84 @@ exclude = '''
...
+
+# Same as Black.
+line-length = 79
+indent-width = 4
+
+target-version = "py38"
+
+[tool.ruff.lint]
+select = ["E4", "E7", "E9", "F", "B", "C4", "UP", "W", "C901"]
```
jendrikseipp, line 59: Please add a comment telling which checks are enabled by ruff by default.

pyproject.toml
```diff
@@ -17,3 +17,84 @@ exclude = '''
...
+ignore = [
+  # C408: unnecessary dict call
+  "C408",
```
jendrikseipp, line 62:
```suggestion
  "C408",  # unnecessary dict call 
```

(use one line)

pyproject.toml
jendrikseipp: Also, please add short comments on what these letters mean. For both of these changes, it's probably best to convert the list to have one entry per line, with a brief comment next to it.

vulture/core.py
```diff
@@ -256,7 +256,7 @@ def handle_syntax_error(e):
             except SyntaxError as err:
                 handle_syntax_error(err)
 
-    def scavenge(self, paths, exclude=None):
+    def scavenge(self, paths, exclude=None):  # noqa: C901
```
jendrikseipp, line 259: Let's just remove C901 from the list of selected tests instead.

tox.ini
```diff
@@ -30,27 +30,20 @@ filterwarnings =
...
+  ruff .
 
 [testenv:fix-style]
 basepython = python3
 deps =
   black==22.3.0
-  pyupgrade==2.28.0
+  ruff==0.1.13
 allowlist_externals =
   bash
```
jendrikseipp, line 46: The bash special case is obsolete now. Same in the "style" part above.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,14 @@
...
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4.1.1
+    - uses: actions/setup-python@v5
```
jendrikseipp, line 13: Is it necessary to set up a Python version here? The ubuntu-latest image already comes with Python installed.

jendrikseipp (changes requested):
Thanks for revising the PR! I have a few more nitpicks, but I like where this is going :-)

I see the differences between ruff and black as an advantage for ruff. So if you like, it would be great if you could change the formatter to ruff in a separate PR later.

Also, feel free to replace the "tox -e style" and "tox -e fix-style" commands with pre-commit in another PR.

Great stuff!

tqa236:
@jendrikseipp I updated the code according to your review. Can you take a look again, please? The list of patterns ignored by `ruff` now is a fuse of `.gitignore` and the one ignored by `black`.

Noted about the future improvement of `ruff format` and `pre-commit`, I can tackle them in follow-up PRs

jendrikseipp:
Great, thanks!

In an upcoming PR, please add a changelog entry for this and the other planned changes. And feel free to add your name or nick to the changelog entry.

PR #361, opened 2024-07-22
Add type hints for the `get_unused_code` function and the fields of `Item`.

johndoknjas:
## Description

This PR adds type hints to make it slightly easier for users to use the api. Now for each element from `get_unused_code`, IDEs will display their type as `Item` when highlighting over the variable. The same is done for the fields of `Item`, such as making `filename`'s type a Path (removing any potential confusion about whether it's a string).

## Checklist:


jendrikseipp:
Thanks for the PR! I'm a bit reluctant to merge it because type annotations without tests might get outdated quickly when the code changes. Is there a way to test that all type annotations are correct, while ignoring all other typing issues that a tool like Mypy might detect? Could you add such a test?

johndoknjas:
@jendrikseipp For sure, I'll work on adding a test to do this.

johndoknjas:
Hey @jendrikseipp, just following up on this PR.

vulture/core.py
```diff
@@ -139,13 +140,13 @@ def __init__(
...
         confidence=DEFAULT_CONFIDENCE,
     ):
-        self.name = name
-        self.typ = typ
-        self.filename = filename
-        self.first_lineno = first_lineno
-        self.last_lineno = last_lineno
-        self.message = message or f"unused {typ} '{name}'"
-        self.confidence = confidence
+        self.name: str = name
```
jendrikseipp, line 143: If I change this line to `self.name: int = name`, all tests still pass. Is it possible to add tests that check that all type hints are followed by the code?

jendrikseipp (changes requested):

vulture/core.py
johndoknjas: @jendrikseipp Sorry about that, the issue was that mypy doesn't know the type of the `name` param, so my test incorrectly still passed. I've updated the test to use the `pytype` linter instead, as it does type inference when hints aren't given.
I noticed that running my test with pytest takes just a few seconds, running with tox is extremely slow; not sure why this is.

reneleonhardt: It's easy to figure out if you take a look at the build log 😅
https://github.com/jendrikseipp/vulture/actions/runs/11144462910/job/30971791967?pr=361#step:8:156

Unfortunately not all dependencies have been updated to support Python 3.13 yet 😢
https://github.com/jcrist/msgspec/issues/698
because the contributed fixes have not been merged yet after three months 🤷
https://github.com/jcrist/msgspec/pull/711
You can see that msgspec doesn't support 3.13:
https://pypi.org/project/msgspec/

An advanced build system like Poetry should check Python support automatically
https://python-poetry.org/docs/dependency-specification/
if Vulture would change its backend:
https://python-poetry.org/docs/basic-usage/

johndoknjas: @reneleonhardt I've updated the PR to only use pytype if python is >= 3.13 (for the time being until it's supported).

jendrikseipp: Hmm, that's unfortunate. I was thinking of switching to `uv`. Would that solve the problem? I won't have time to look into this for the next weeks, though.

reneleonhardt: Nice solution!
I hope msgspec and pytype will support 3.13 soon as it will be released on Monday (but identical to rc3)...

tests/test_pytype.py
```diff
@@ -0,0 +1,15 @@
...
+@pytest.mark.skipif(
+    sys.version_info >= (3, 13), reason="Need python < 3.13 for pytype"
+)
+def test_pytype():
+    result = subprocess.run(
+        ["pytype", "vulture/core.py", "--disable=attribute-error"],
+        capture_output=True,
+        text=True,
+    )
+    assert result.stdout.strip().endswith("Success: no errors found")
```
jendrikseipp, line 15: Let's check whether the exitcode is 0 to make this more robust.

tests/test_pytype.py
```diff
@@ -0,0 +1,15 @@
+import subprocess
+import sys
+import pytest
+
+
+@pytest.mark.skipif(
+    sys.version_info >= (3, 13), reason="Need python < 3.13 for pytype"
```
jendrikseipp, line 7:
```suggestion
    sys.version_info >= (3, 13), reason="needs Python < 3.13 for pytype"
```

tests/test_pytype.py
```diff
@@ -0,0 +1,15 @@
+import subprocess
+import sys
+import pytest
```
jendrikseipp, line 3: Add empty line above pytest import.

tests/test_pytype.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+@pytest.mark.skipif(
+    sys.version_info >= (3, 13), reason="Need python < 3.13 for pytype"
+)
+def test_pytype():
```
jendrikseipp, line 9: I think ideally we would just call pytype directly from tox.ini, but I don't know whether it's possible to only run the test for Python < 3.13. Can you check?

tests/test_pytype.py
```diff
@@ -0,0 +1,15 @@
...
+    result = subprocess.run(
+        ["pytype", "vulture/core.py", "--disable=attribute-error"],
```
jendrikseipp, line 11: Instead of disabling the error, let's define `self.noqa_lines = {}` in the constructor alongside the other member variables.

jendrikseipp (changes requested):
Thanks for your persistence :)

tests/test_pytype.py
johndoknjas: I researched around and tried a few things, but nothing seemed to work. I feel there might be a way, but it may not really be supported by tox's default behaviour. I suppose once pytype and msgspec support 3.13, we could switch to a direct call in tox?

jendrikseipp: Thanks for trying! I agree that using pytest to call pytype for now is the best option.

tox.ini
```diff
@@ -18,7 +18,7 @@ deps =
   pytype ; python_version < '3.13'
 commands =
   pytest {posargs}
-  ["$PYTHON_VERSION" -lt "3.13"] && pytype vulture/core.py
+  py{38,310,311,312}: pytype vulture/core.py
```
reneleonhardt: Was there a specific reason to exclude 39?
3.8 will be EOL in a month, but 3.9 in a year 😅

jendrikseipp: For some reason tox stumbles over py39. Python 3.9 is sill tested via GitHub Actions.

tests/test_pytype.py
johndoknjas: @jendrikseipp Oh the `check=True` does that. If the exitcode isn't 0, subprocess will raise a CalledProcessError.

jendrikseipp: I think it's best to make that check explicit and to remove the string comparison (and to remove the then obsolete kwargs from the run call).

jendrikseipp:
Thanks for your work on this!

johndoknjas:
No problem, and thank you for your feedback :)

PR #376, opened 2024-11-18
Remove `pass_filenames: false` from pre-commit hook

cbachhuber:
To make the vulture pre-commit hook work on the specific files specified by pre-commit and not on the entire repo, I need to override the config it as shown [here](https://github.com/luminartech/dev-tools/pull/43/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R159), with `pass_filenames: true`. I argue that this is a saner default than `pass_filenames: false`, so let's remove it from vulture's pre-commit hook.

## Related Issue

https://github.com/luminartech/dev-tools/pull/43

## Checklist:

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,6 +4,5 @@
   entry: vulture
   description: Find unused Python code.
   types: [python]
-  pass_filenames: false
   require_serial: true
   


hofbi (approved):

cbachhuber:
That makes sense, thanks for explaining! How about adding an explanatory comment to this config line? Something like

```yaml
--pass-filenames: false # vulture needs access to all files for a complete analysis
```

Btw, in our private repo, we're using `--pass-filenames: true` because it's a large legacy codebase and we haven't rolled out vulture completely yet. Together with `min_confidence = 100`, it works just fine. With this setup, the value of vulture is very limited, of course.

jendrikseipp:
Good idea about the comment. Done.

PR #378, opened 2024-11-28
Handle while True loops without break statements

kreathon:
## Description

In the following example, the `print` statement is not recognized as unreachable code by `vulture`. 

```
while True:
    raise Exception()
print("Currently not recognized!")
```

More general: If a `while True` statement has no `break`, then the statement after the `while True` is unreachable.

## Implementation

Introduce `self._current_loop_has_break_statement` in order to keep track of `break` statements found while traversing the AST. 

We cannot validate if the break statements only occur in loops (so we need to assume the code is correct). In order to check this properly, we would need to visit the parent first (and push some kind of context or scope; pop if after visiting the children).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # next (unreleased)
 
+* Handle `while True` loops without `break` statements (kreathon).
 * Improve reachability analysis (kreathon, #270, #302).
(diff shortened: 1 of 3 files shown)


jendrikseipp:
The code is flawless. Thanks for the PR!

Issue #216, opened 2020-07-22
Support Forward References for Annotations


colinc719:
When vulture is run on the code

```python
from collections import Sequence


def temp(arg: "Sequence"):
    assert not arg


if __name__ == '__main__':
    temp([])
```

produces "unused import 'Sequence'"

I would expect it to see that Sequence is used as a forward reference (https://www.python.org/dev/peps/pep-0484/#forward-references) in temp.

I'm running vulture 1.5.

jendrikseipp:
Thanks for the report! Vulture currently doesn't parse string forward references, but I guess it would make for a nice addition. In your concrete example, however, couldn't you just use `Sequence` instead of the string `"Sequence"`?

colinc719:
Thanks for the fast response.  Yeah, I was just trying to post a simple example.  A more realistic use-case would probably use TYPE_CHECKING, which gives the same result:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from collections import Sequence


def temp(arg: "Sequence"):
    assert not arg


if __name__ == '__main__':
    temp([])
```

jendrikseipp:
Thanks for the example!

jendrikseipp:
#220 added support for parsing type comments, but forward references are still not taken into account by Vulture.

jingw:
Just FYI, fixing this fully correctly is likely annoying, since the forward references can appear outside a type annotation context:
```
MyType = Union["LaterType", Whatever]

class LaterType:
    pass
```

That said, somehow flake8 deals with it, so it could be worth looking at whatever they're doing. Test case:
```
from foobar import Thing
from typing import Union

Thing2 = Union["Thing", int]
```

(not planning to work on this, just throwing this out)

jendrikseipp:
Thanks for the explanation!

Maybe a solution could be to record all literal strings as used names. (This would render the getattr/hasattr PR obsolete.) My feeling is that this sounds too drastic, but could actually be the right move. On the other hand, maybe we shouldn't worry about this too much, given that there's `from __future__ import annotations` (thanks for the hint, didn't know about this!). What do you think?

I also thought about dropping unused import detection, but I think Vulture actually has some value in addition to flake8 for detecting unused imports:

``` python
# a.py
import sys

# b.py
import a
print(a.sys.argv)
```

Flake8 reports `sys` as an unused import, but Vulture doesn't.

jingw:
I also feel that would be excessive.

[...]

PR #379, opened 2024-12-15
Add forward reference annotation detection

Joko013:
Adding support for parsing forward references in type annotations and using them for the detection algorithm.

## Description
Hi, I'd like to add the support for [forward references](https://peps.python.org/pep-0484/#forward-references) so I would appreciate it if you could take a look at this draft.

I understand that [it has been discussed](https://github.com/jendrikseipp/vulture/tree/main?tab=readme-ov-file#forward-references-for-type-annotations) but I think it's worth reconsidering, given the timeframe (the original discussion is from 2020) and further adoption of new Python versions.

Personally, this blocks the usage of Vulture as it either forces you to adopt a new standard (`from __future__ import annotations` instead of `if TYPE_CHECKING:...`) or deal with a lot of false positives if you don't (or can't).

I know that in the original discussion, there's been concerns about the issues that this could open up so I've tried to cover some basic cases but I'm all for for covering more.

Let me know if this approach makes sense to you. 

## Related Issue
- https://github.com/jendrikseipp/vulture/issues/216

## Checklist:

### DIFF
--- a/tests/test_annotations_parser.py
+++ b/tests/test_annotations_parser.py
@@ -1,0 +1,22 @@
+import typing
+
+import pytest
+
+from vulture.annotation_parser import AnnotationParser
+
+
+@pytest.mark.parametrize(
+    "input_annotation, expected_types",
+    [
+        ("Foo", {"Foo"}),
+        ("foo.Foo", {"foo", "Foo"}),
+        ("List['Foo', 'Bar']", {"List", "Foo", "Bar"}),
+        ('List["Foo", "Bar"]', {"List", "Foo", "Bar"}),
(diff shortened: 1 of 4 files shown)


jendrikseipp:
Thanks for the PR! I'll take a look when I find the time.

thearchitector:
I'm quite interested in supporting forward references as well.

I'm also wondering if this may be made easier, or at least affected, by 3.14's default [deferring of annotations](https://docs.python.org/3.14/whatsnew/3.14.html#whatsnew314-deferred-annotations), as well as its new [annotationlib](https://docs.python.org/3/library/annotationlib.html)? The latter has backports via `typing-extensions`.

PR #400, opened 2025-09-30
Add optional --header argument to report

tadeubas:
Add optional `--header` option to Vulture report, specially useful when generating whitelist files.

## Description

This PR introduces a new option, `--header`, which prints a header line before listing the first unused item.

The header follows the format:
```bash
# Vulture v2.14 – Find dead code
```

## Checklist:


jendrikseipp:
Thanks for the PR, but I think it's best to keep the output as terse as possible.

tadeubas:
Yes, that’s why the parameter is optional.

Issue #407, opened 2025-11-25
False negative: Method calling another method with same name not detected as unused


ronald-braunstein:
# False negative: Method calling another method with same name not detected as unused

## Description

Vulture fails to detect unused methods when a method calls another method with the **same name**, even though nothing calls the wrapper method.

## Environment

- Vulture version: 2.7
- Python version: 3.10

## Minimal Reproducible Example

```python
# test_same_name_bug.py
class InternalClass:
    @staticmethod
    def same_name_method():
        return "internal"

class PublicClass:
    @staticmethod
    def same_name_method():
        # This wrapper is NEVER called, but vulture doesn't report it
        return InternalClass.same_name_method()

    @staticmethod
    def different_name_wrapper():
        # This wrapper is also never called, and vulture DOES report it
        return InternalClass.same_name_method()

    @staticmethod
    def used_method():
        return "I am used"

# Use PublicClass so it's not marked as totally unused
result = PublicClass.used_method()
```

## Expected Behavior

Running `vulture test_same_name_bug.py --min-confidence 0` should report:
```
test_same_name_bug.py:9: unused method 'same_name_method' (60% confidence)
test_same_name_bug.py:14: unused method 'different_name_wrapper' (60% confidence)
```

## Actual Behavior

Vulture only reports:
```
test_same_name_bug.py:24: unused variable 'result' (60% confidence, 1 line)
test_same_name_bug.py:14: unused method 'different_name_wrapper' (60% confidence, 4 lines)
```

The `PublicClass.same_name_method` at line 9 is **NOT** reported as unused.

## Root Cause

When a method makes a call to `OtherClass.same_name_method()`, vulture appears to mark **both** the caller and callee as "used" if they share the same method name, even though nothing actually calls the wrapper method.

## Real-World Impact

This bug was discovered in production code where we have a service layer that wraps internal methods:

```python
class FileReader:
    @staticmethod
    def deprecated_get_file_by_file_path_optional(ctx, *, file_path: str):
        # Internal implementation
        ...

class CloudStorageService:
    @staticmethod
    def deprecated_get_file_by_file_path_optional(ctx, *, file_path: str):
        # This wrapper is never called anywhere in the codebase
        # But vulture doesn't report it because it calls FileReader's method with the same name
        return FileReader.deprecated_get_file_by_file_path_optional(ctx, file_path=file_path)
```

[...]

Issue #371, opened 2024-10-05
False negative when method and function have the same name


johndoknjas:
```
def f():
    pass

class F:
    def f():
        pass

F.f()
```
Vulture doesn't give any errors here. It seems to be because the ids are the same.

TravisDart:
Here's a simpler example. The program below doesn't detect any unused code:

```
def hello_world1():
    message = "Hello, world!"  # Unused
    print("Hello, world!")

def hello_world2():
    message = "Hello, world!"
    print(message)

if __name__ == "__main__":
    hello_world1()
    hello_world2()
```

However, the code below returns: `unused variable 'message' (60% confidence)`

```
def hello_world1():
    message = "Hello, world!"  # Unused
    print("Hello, world!")

def hello_world2():
    message2 = "Hello, world!"  # Renamed to message2
    print(message2)

if __name__ == "__main__":
    hello_world1()
    hello_world2()
```

Issue #366, opened 2024-09-25
"unused variable" does not appear when variable with same name in other file 


dpinol:
It looks like running vulture with confidence=100, is not reporting  "unused variable" for function arguments named as other variables in other files. However, this does not happen if the "other variable" occurs in the same file.

## kk.py
```kk.py
def kk():
    expr:int =3
 
def fn2(expr: int, expr2: int) -> None:
    pass
 
def qq():
    expr:int =3

kk()
fn2()
qq()
```
### qq.py
```
def io():
    expr2=3
    return expr2

io()
```
```bash
vulture --min-confidence=10 kk.py qq.py
kk.py:3: unused variable 'expr' (60% confidence)
kk.py:5: unused variable 'expr' (100% confidence)
kk.py:9: unused variable 'expr' (60% confidence)
```

 

thanks

Issue #343, opened 2023-12-06
Feature request: A way to more specifically target items for whitelisting


Hubro:
I'm trying to implement Vulture for a big project at work that uses a proprietary application framework.

Entrypoints into the application are classes with a specific name (`Main`) that are subclasses of a specific framework class (`acme_corp.application.Service`) placed in a very specifically named module (e.g. `my_package_name/main.py`). The framework also has hooks, which are defined as methods with very specific names in subclasses of a specific framework class.

This is a very large codebase, and simply whitelisting the names `Main`, `_.before`, `_.after` etc. is likely to hit much more than just the entrypoints that I want to whitelist.

It would be tremendously useful if Vulture leveraged the AST to allow even more specific whitelists. You could take into account the type of object (if it's a constant, variable, class, method, function etc.), the location (file name, module name, package name), the base classes and so on.

To maintain backwards compatability, this could be added as a new feature, for example:

**pyproject.toml**:

```toml
[tool.vulture]

paths = "packages/"

entrypoints = [
    # Any class named "Main" in a module whose name ends with ".main" that is a
    # subclass of "acme_corp.application.Application"
    {
        type = "class",
        name = "Main",
        base_classes = ["acme_corp.application.Application"],
        module_name = "/\\.main$/",
    },

    # Any method named "create", "pre_hook" or "post_hook" in a class like the
    # one above
    {
        type = "method",
        name = ["create", "pre_hook", "post_hook"],
        parent_class = {
            name = "Main",
            base_classes = ["acme_corp.application.Application"],
            module_name = "/\\.main$/",
        }
    },

    # Any method named "pre_hook" or "post_hook" in a subclass of
    # "acme_corp.application.Service". (The class itself is *not* whitelisted.)
    {
        type = "method",
        name = ["pre_hook", "post_hook"],
        parent_class = {
            base_classes = ["acme_corp.application.Service"],
        }
    },

    # Class attributes named "name_prefix" in subclasses of
    # "acme_corp.application.Service". (The class itself is *not* whitelisted.)
    {
        type = "class_attribute",
        name = "name_prefix",
        parent_class = {
            base_classes = ["acme_corp.application.Service"],
        }
    },
]
```

Alternative names: `mark_used`, `tag`, (?)

Amvoled:
I've started using vulture a few days ago and I've run into the same problem. Sometimes we just implement new features incrementally and therefore leave section of code unused for now with the intent of using them later.
[...]

PR #417, opened 2026-05-17
Distinguish same-named functions across modules

ethanleifer:
## Summary

This changes function tracking so functions with the same local name in different modules are treated as distinct definitions when Vulture can derive a module from the scanned filename.

Today, Vulture stores function definitions by bare name only. That means a use of `load_config` in one module can mark every `load_config` definition in the scanned project as used, including definitions in unrelated modules. This creates false negatives for projects that commonly reuse conventional function names across modules.

## Motivation

For example, with two modules that each define a `load_config` function, calling one of them should not suppress the unused report for the other:

```python
# package/commands/loaders.py
def load_config():
    pass

# package/reports/loaders.py
def load_config():
    pass

# package/main.py
from package.commands.loaders import load_config

load_config()
```

After this change, Vulture reports `package.reports.loaders.load_config` as unused while recognizing `package.commands.loaders.load_config` as used.

## Approach

- Derive a module name from `.py` filenames, including `__init__.py` handling.
- Keep `Item.name` as the local identifier, e.g. `load_config`.
- Store the module separately on `Item.module`, e.g. `package.commands.loaders`.
- Expose `Item.full_name` for function matching/reporting, e.g. `package.commands.loaders.load_config`.
- Preserve the existing bare-name `used_names` set for current behavior and public test expectations.
- Add an internal full-name usage set populated from direct, relative, and module attribute imports.
- Keep classes, methods, properties, variables, attributes, and imports on their existing naming behavior.

## Compatibility notes

This changes report and whitelist output for unused functions when Vulture is given filenames that can be converted to module names. For example, `foo.py` now reports `foo.myfunc` instead of `myfunc`.

I kept the change scoped to functions to address the cross-module false-negative case without attempting full scope-aware analysis. Methods with the same name on different classes are still intentionally outside the scope of this PR.

## Related issues

This is related to the broader set of false negatives caused by name-only matching, including:

- #366
- #371
- #407
- #343

## Tests

- `python -m ruff format tests/test_item.py tests/test_scavenging.py`
- `python -m ruff check vulture/core.py tests/test_item.py tests/test_scavenging.py tests/test_report.py`
- `python -m pytest tests/test_item.py tests/test_scavenging.py tests/test_report.py`
- `python -m pytest`
- `python -m coverage report -m vulture/core.py`


mcdigman:
Hi! 

This commit changes a large number of findings, (mostly just changing the names of symbols flagged, but some others), and I think not all the changes are intended.

Here are consolidated review findings from the comments made in [a simulated version of this PR on my personal fork](https://github.com/mcdigman/vulture/pull/9). I used this PR as a demo example of a PR with a large blast radius in Issue #431 (This PR produced 993 new and 764 dropped findings across 24 projects in the corpus of GitHub repositories I scanned)

1. For functions with `Item.module`, `_get_unused_funcs` no longer consults `used_names`. A use suppresses a finding only when `vulture` associates it with the same full name in `used_full_names`. Imports, local-definition bindings, and an encounter-order rescue can establish that association, but dynamic attribute dispatch, `getattr`, decorator registries, and similar bare-name evidence generally cannot. Those uses therefore become new false positives.

2. Module identity comes from the filesystem path, not the import graph. This is the most concrete issue. `_module_name_from_filename` yields `src.black.strings` while the import binding says `black.strings`, so in a `src`-layout project *any* function used only cross-module and defined before its use is a guaranteed false positive. Also current working directory-dependent: running from a parent directory produced `repro3.src.pkg.a_defs.helper`, and an absolute path outside cwd hits the `suppress(ValueError)` and turns the whole filesystem path into the module name.

3. `name_bindings` conflates imports and definitions in one flat dict. `_record_name_binding` overwrites; `_define` uses `setdefault`. A module that both imports and defines a name gets one binding and the loser is reported unused — breaking the standard conditional-fallback pattern in both directions. Real examples: `try/except ImportError` (`more_itertools/more.py`:210 `_fsumprod`, `httpx/__init__.py`:18 `main`, `src/requests/adapters.py`:66 `SOCKSProxyManager`) and `sys.platform` branches (`src/platformdirs/unix.py`:21 `getuid`).

4. Uses textually preceding their binding site record no full name. Real examples: `src/click/_winconsole.py`:277 `_get_windows_console_stream` is called at `_compat.py`:341/348/355, before the lazy `from ._winconsole import …` at line 518. Common with deferred imports.

5. Scan order is filesystem order, so results are not reproducible across machines. `vulture -v` on `black` yields `files.py`, `parsing.py`, `handle_ipynb_magics.py`, `rusty.py`, `concurrency.py`…. The `if name` in `self.used_names` rescue in `_define` fires or not depending on directory iteration order — `src/black/strings.py`:367 `str_width` is flagged in the CI run but not in a local run of the same commit against the same source.

6. Methods and dunders are module-qualified despite the PR excluding methods. Vulture types implicit classmethods and protocol methods as function, so _define qualifies them: my analysis had ~21 such new findings, including `pydantic.main.__class_getitem__`, `pydantic.main.__init_subclass__`, `src.flask.app.__init_subclass__`, `xarray…__array__`. The resulting name is also wrong — it omits the class. Same for ordinary methods (`starlette.requests.url_for`, `pydantic.type_adapter.dump_python`). Fix: qualify only module-level functions.

7. Whitelist compatibility is broken for affected functions. Existing bare-name entries no longer suppress module-qualified findings, while `--make-whitelist` emits module.func; rescanning that expression normally records only the bare attribute name, so the newly generated whitelist also fails to suppress its source finding. Base round-trips successfully, while head continues to report `foo.other`. Thus users can neither retain the affected old entries nor reliably regenerate replacements.

8. `__all__` listed below the definition is a new false positive:
```
def exported():
    pass

__all__ = ["exported"]
```

PR #426, opened 2026-07-21
Detect single-character __all__ string exports

### LINKED ISSUE #425: Consider `__all__` when it is a string with a single character
```python
__all__ = 'a'
a = None # flagged
```

```python
__all__ = 'a',
a = None # not flagged
```

But the two are functionally equivalent. `str` has all of `tuple`'s methods, and in the single-character case, iteration, indexing, etc. behave identically.

Phoenix0531-sudo:
## Summary

`__all__ = "a"` does not mark `a` as used today, while `__all__ = ("a",)` does. For a single-character string those forms are equivalent when iterating export names, so vulture was raising a false positive on the string form (issue #425).

This change only special-cases **length-1** string constants assigned to `__all__`. Multi-character strings such as `__all__ = "Foo"` are left alone so existing behavior and tests stay intact.

## Testing

- Added unit tests for single-character `__all__` with imports/variables (and a tuple regression)
- `pytest` green locally for the suite relevant to this change (299 passed; one pre-existing env failure on `pint` whitelist import when `pint` is not installed)

## Related

Fixes #425

Note: open PR #421 improves other `__all__` assignment forms (`+=`, annotated). This patch is orthogonal and only covers the single-character string case on plain assignment.


Phoenix0531-sudo:
Friendly ping — this small fix for issue #425 is still relevant on current main.
All checks pass locally and in CI. Happy to adjust anything if needed.

Phoenix0531-sudo:
The maintainer was active on Sep 16 (merging #428/#429) — just noting this CI-green PR is still open and applies cleanly to current main.

vulture/core.py
```diff
@@ -617,6 +617,15 @@ def visit_Assign(self, node):
...
                     self.used_names.add(elt.value)
+        elif any(
+            isinstance(target, ast.Name) and target.id == "__all__"
+            for target in node.targets
+        ) and utils.is_ast_string(node.value):
+            # `__all__ = "a"` is equivalent to `__all__ = ("a",)` for a
+            # single-character name (issue #425). Multi-character strings are
+            # intentionally ignored (see test_import_with__all__string).
+            if len(node.value.value) == 1:
+                self.used_names.add(node.value.value)
```
jendrikseipp, line 627: The code shouldn't duplicate the functionality of `_assigns_special_variable__all__`.

jendrikseipp (changes requested):

PR #428, opened 2026-08-15
Report an out-of-range --min-confidence as a usage error, not a traceback

VXNCXNX:
## Description

An out-of-range `--min-confidence` exits with a Python traceback rather than a usage error.

```
$ vulture --min-confidence 150 a.py
Traceback (most recent call last):
  ...
  File "vulture/core.py", line 322, in get_unused_code
    raise ValueError("min_confidence must be between 0 and 100.")
ValueError: min_confidence must be between 0 and 100.
```

After:

```
$ vulture --min-confidence 150 a.py
min_confidence must be between 0 and 100
```

`--help` documents the flag as "Minimum confidence (between 0 and 100)", so the value is already specified; the check just lived in the wrong place. It sat in `get_unused_code()`, which runs after the files have been scanned, and raised a bare `ValueError` that nothing catches.

Moving it into `_check_output_config` puts it next to the existing "Please pass at least one file or directory" check, so it raises `InputError` and goes through the same path as every other bad-argument case. That also means the exit code is now `InvalidCmdlineArguments` (2) rather than `InvalidInput` (1), which matches what the other argument errors return.

Both ends are covered: `--min-confidence -5` gives the same error, while `0` and `100` are still accepted.

## Related Issue

None; I did not find a matching open issue.

## Checklist:

- [x] I have updated the documentation in the README.md file or my changes don't require an update. (`--help` already documents the 0-100 range, this makes the code enforce it in the right place.)
- [x] I have run `pre-commit run --all-files` to format and lint my code. (`ruff format` Passed, `ruff check` Passed.)

## Testing

`test_min_confidence_out_of_range` in `tests/test_config.py`, parametrised over `-1` and `101`, next to the existing `test_missing_paths` which covers the other `_check_output_config` branch.

Reverting the check fails both parameters:

```
Failed: DID NOT RAISE InputError
2 failed, 18 passed
```

`pytest tests/test_config.py` is 20 passed.

*Disclosure: written with AI assistance (Claude Code). I ran the before and after above and the revert check myself.*


jendrikseipp:
Thanks!

PR #429, opened 2026-08-15
Ignore empty --exclude patterns instead of excluding every file

VXNCXNX:
## Description

An empty `--exclude` pattern excludes every file, so vulture reports nothing and exits 0.

A trailing comma is enough to trigger it:

```
$ cat a.py
import os

$ vulture --exclude=zzz, a.py
$ echo $?
0
```

No output, no error, exit 0. With `-v` the cause is visible:

```
Excluded: /tmp/a.py
```

After:

```
$ vulture --exclude=zzz, a.py
a.py:1: unused import 'os' (90% confidence)
$ echo $?
3
```

The same happens from `pyproject.toml` with `exclude = ["zzz", ""]`.

`--exclude` splits its value on `,` with no filtering, so a trailing comma yields an empty string. `prepare_pattern` then wraps anything without a `*` as `f"*{pattern}*"`, turning `""` into `**`, which matches every path.

This is a bad failure mode for a linter specifically: a clean run and exit 0 are indistinguishable from success, so it silently stops checking anything, in CI as much as locally.

## The fix

Skip empty patterns when building the exclude list. A real pattern is unaffected: `--exclude=a.py a.py` still excludes.

## Related Issue

None; I did not find a matching open issue.

## Checklist:

- [x] I have updated the documentation in the README.md file or my changes don't require an update. (No doc change: an empty pattern was never a documented way to exclude everything.)
- [x] I have run `pre-commit run --all-files` to format and lint my code. (`ruff format` Passed, `ruff check` Passed.)

## Testing

`test_exclude_empty_pattern` in `tests/test_script.py` scavenges a module with `exclude=["", "no_match"]` and asserts the unused import is still reported.

Reverting `core.py` fails it:

```
assert [item.name for item in v.get_unused_code()] == ["os"]
E   AssertionError: assert [] == ['os']
```

`pytest tests/test_script.py` is 13 passed.

*Disclosure: written with AI assistance (Claude Code). I ran the before and after above and the revert check myself.*


VXNCXNX:
Both of these have been open a month, both green and level with `main`, so they are only waiting on a look.

Two independent argument-validation fixes:

- **#428**, `+13/-0`: `--min-confidence 150` exits with a Python traceback rather than a usage error. It now reports the range the way argparse reports every other bad value.
- **#429**, `+16/-2`: an empty `--exclude` pattern excludes every file, so vulture reports nothing and exits 0.

#429 is the one with a real failure mode behind it, and a trailing comma is enough to trigger it:

```
$ vulture --exclude=zzz, a.py
$ echo $?
0
```

No output, no error, exit 0, so a scan that silently covers nothing still looks like a clean run.

Happy to rebase, or to close either if you would rather argparse stayed untouched here.

